PR #7, opened 2022-02-17
Inital core implementation

MagicRB:


lib/Entry.hs
```diff
@@ -0,0 +1,115 @@
...
+instance Hashable FieldKey
+
+instance A.ToJSON FieldKey where
+instance A.ToJSONKey FieldKey where
+instance A.FromJSON FieldKey where
+instance A.FromJSONKey FieldKey where
+
+newFieldKey :: T.Text -> Maybe FieldKey
+newFieldKey t =
+  if T.foldr ((&&) . (`elem` allowedChars)) True t then
```
dcastro, line 39:
```suggestion
  if T.all (`elem` allowedChars) t then
```

lib/Entry.hs
```diff
@@ -0,0 +1,115 @@
...
+
+newtype EntryTag = EntryTag T.Text
+  deriving (Generic, Show, Eq)
+
+instance A.ToJSON EntryTag where
+instance A.FromJSON EntryTag where
+
+newEntryTag :: T.Text -> Maybe EntryTag
+newEntryTag t =
+  if T.foldr ((&&) . (`elem` allowedChars)) True t then
```
dcastro, line 56:
```suggestion
  if T.all (`elem` allowedChars) t then
```

lib/Entry.hs
```diff
@@ -0,0 +1,115 @@
...
+
+instance Hashable FieldKey
+
+instance A.ToJSON FieldKey where
+instance A.ToJSONKey FieldKey where
+instance A.FromJSON FieldKey where
+instance A.FromJSONKey FieldKey where
+
+newFieldKey :: T.Text -> Maybe FieldKey
+newFieldKey t =
```
dcastro, line 38: We should also check that the input text has at least 1 character.

And because now we have two possible reasons for `newFieldKey` to fail, it should return an `Either Text FieldKey`, where `Text` contains the reason why it failed.

This also goes for `newEntryTag`

lib/Entry.hs
```diff
@@ -0,0 +1,115 @@
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+module Entry
+  ( dateModified
+  , Entry, EntryConvertible (..), newEntry
+  , path, masterField, fields
+  , Field (..), FieldKey (..)
```
dcastro, line 8:
```hs
FieldKey (..)
```

The field `FieldKey` has some invariants (all characters are valid & has at least 1 character), so its default constructor `FieldKey :: Text -> FieldKey` should not be exported.

lib/Entry.hs
```diff
@@ -0,0 +1,115 @@
...
+  , newField, newFieldKey, getFieldKey, newEntryTag, getEntryTag
+  , private, value, tags
+  )
```
dcastro, line 11: The `EntryTag` _type_ should be exported (but not its constructor).

lib/Entry.hs
```diff
@@ -0,0 +1,115 @@
...
+import Control.Lens
+import qualified Data.Aeson.Types as A
+import Data.Hashable (Hashable)
+import GHC.Generics (Generic)
+import Data.Time (UTCTime)
+import Language.Haskell.TH.Lens (_CaseE)
+
+type DateTime = UTCTime
+
+newtype FieldKey = FieldKey T.Text
```
dcastro, line 27: When a data type holds internal invariants (all characters must be valid & has at least 1 character) and is supposed to be constructed via a "smart constructor" (like `newFieldKey`), it's customary to add the prefix `Unsafe` to the constructor name.

For example, here's a quick query of the [morley codebase](https://gitlab.com/morley-framework/morley).
image

This helps draw attention to the fact that usage of this constructor is discouraged and, when used, should be used with care.

```suggestion
newtype FieldKey = UnsafeFieldKey T.Text
```

lib/Entry.hs
```diff
@@ -0,0 +1,115 @@
...
+  where allowedChars = ['a'..'z'] ++ ['A'..'Z'] ++ ['0'..'9'] ++ "-_;:"
+
+getFieldKey :: FieldKey -> T.Text
+getFieldKey (FieldKey t) = t
+
+newtype EntryTag = EntryTag T.Text
+  deriving (Generic, Show, Eq)
+
+instance A.ToJSON EntryTag where
+instance A.FromJSON EntryTag where
```
dcastro, line 52: `DerivingStrategies` + `GeneralizedNewtypeDeriving` really helps here.

```suggestion
newtype EntryTag = EntryTag T.Text
  deriving stock (Show, Eq)
  deriving newtype (A.ToJSON, A.FromJSON)
```

lib/Entry.hs
```diff
@@ -0,0 +1,115 @@
...
+
+newtype FieldKey = FieldKey T.Text
+  deriving (Generic, Show, Eq)
+
+instance Hashable FieldKey
+
+instance A.ToJSON FieldKey where
+instance A.ToJSONKey FieldKey where
+instance A.FromJSON FieldKey where
+instance A.FromJSONKey FieldKey where
```
dcastro, line 35:
```suggestion
  deriving stock (Show, Eq)
  deriving newtype (A.ToJSON, A.ToJSONKey, A.FromJSON, A.FromJSONKey, Hashable)
```

lib/Entry.hs
```diff
@@ -0,0 +1,115 @@
...
+
+newFieldKey :: T.Text -> Maybe FieldKey
+newFieldKey t =
+  if T.foldr ((&&) . (`elem` allowedChars)) True t then
+    Just $ FieldKey t
+  else
+    Nothing
+  where allowedChars = ['a'..'z'] ++ ['A'..'Z'] ++ ['0'..'9'] ++ "-_;:"
```
dcastro, line 43: As discussed [here](https://serokell.slack.com/archives/C024MA7K48Z/p1642788097019600?thread_ts=1642295588.004200&cid=C024MA7K48Z), can you please disallow the character `:` from appearing in field names?

lib/Entry.hs
```diff
@@ -0,0 +1,115 @@
...
+instance A.ToJSON EntryTag where
+instance A.FromJSON EntryTag where
+
+newEntryTag :: T.Text -> Maybe EntryTag
+newEntryTag t =
+  if T.foldr ((&&) . (`elem` allowedChars)) True t then
+    Just $ EntryTag t
+  else
+    Nothing
+  where allowedChars = ['a'..'z'] ++ ['A'..'Z'] ++ ['0'..'9'] ++ "-"
```
dcastro, line 60: I see no reason for tags and field names to allow different character sets. Can we use the same set for both?

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,234 @@
...
+        exceptionHandler :: [T.Text] -> ClientError -> Maybe CofferError
+        exceptionHandler path =
+          \case FailureResponse _request response ->
+                    case statusCode $ responseStatusCode response of
+                      404 -> Nothing
+                      e -> Just $ OtherError (T.pack $ show e)
+                DecodeFailure text response -> Just MarshallingFailed
+                UnsupportedContentType mediaType response -> Just MarshallingFailed
+                InvalidContentTypeHeader response -> Just MarshallingFailed
+                ConnectionError excepton -> Just ConnectError
```
dcastro, line 196:
```suggestion
                ConnectionError exception -> Just ConnectError
```

lib/Entry.hs
```diff
@@ -0,0 +1,115 @@
...
+
+makeLensesFor [("_value", "value"), ("_private", "private")] ''Field
+
+data Entry =
+  Entry
+  { _path :: [T.Text]
+  , _eDateModified :: DateTime
+  , _masterField :: Maybe FieldKey
+  , _fields :: HS.HashMap FieldKey Field
+  , _tags :: [EntryTag]
```
dcastro, line 89: Tags should be in a set, rather than a list, to prevent duplicates.

flake.nix
```diff
@@ -0,0 +1,37 @@
+{
```
notgne2, line 1: I think this flake in general probably needs cleaning up

DK318 (commented):
Everything seems ok for me, except for open conversations.

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,234 @@
...
+  where tokenToText (I.VaultToken t) = Just t
+        textToToken t = Just $ I.VaultToken t
+        baseUrlToText = Just . T.pack . showBaseUrl
+        textToBaseUrl = parseBaseUrl . T.unpack
+
+data FieldMetadata = FieldMetadata
+  { _dateModified :: UTCTime
+  , _private :: Bool
+  }
+  deriving (Show, Generic)
```
Heimdell, line 85:
```suggestion
  deriving stock (Show, Generic)
  deriving anyclass (A.ToJSON, A.FromJSON)
```

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,234 @@
...
+        (BaseUrl Http _ _ _) -> do
+          manager <- embed $ newManager defaultManagerSettings
+          pure $ mkClientEnv manager url
+        (BaseUrl Https _ _ _) -> do
+          manager <- embed $ newManager tlsManagerSettings
+          pure $ mkClientEnv manager url
+
+    case x of
+      WriteSecret entry -> do
+        let cofferSpecials = CofferSpecials
```
Heimdell, line 124: You can do like
```
let
  cofferSpecials = CofferSpecials
    { ...
    , ...
    }
  secret = I.PostSecret
    { ...
    }
```
This way the indent won't be so ridiculos.

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,234 @@
...
+
+          pure $ E.newEntry path (cofferSpecials ^. globalDateModified)
+            & E.masterField .~ (cofferSpecials ^. masterKey >>= E.newFieldKey)
+            & E.fields .~ fields
+            & E.tags .~ _tags
+      ListSecrets path ->
+        runMaybe (embedCatchMaybe path (listSecrets env path) <&> (^. I.ddata) <&> \(I.ListSecrets list) -> list)
+      DeleteSecret path ->
+        embedCatch path (deleteSecret env path) <&> const ()
+  where postSecret env = (I.routes env ^. I.postSecret) mount token
```
Heimdell, line 181: 1) We have styleguide; 
2) that `where` is a major offence to the reader.

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,234 @@
...
+        readSecret env = (I.routes env ^. I.readSecret) mount token
+        listSecrets env = (I.routes env ^. I.listSecrets) mount token
+        updateMetadata env = (I.routes env ^. I.updateMetadata) mount token
+        deleteSecret env = (I.routes env ^. I.deleteSecret) mount token
```
Heimdell, line 185: These four should be one
```hs
perform env act = (I.routes env ^. act) mount token
```

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,234 @@
...
+        runMaybe :: Sem (Error () ': r) t
+                 -> Sem r (Maybe t)
+        runMaybe x = runError x <&> \case
+                       Left _ -> Nothing
+                       Right a -> Just a
+
+        maybeThrow :: Member (Error CofferError) r
+                      => Maybe a
+                      -> Sem r a
+        maybeThrow = maybe (throw MarshallingFailed) pure
```
Heimdell, line 227: There 4 should be in some `Util.hs`.

Why a function that is called `maybeThrow` throws domain-related exception? Should be abstracted out.

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,342 @@
...
+data KvResponse a =
+  KvResponse
+  { _requestId :: T.Text
+  , _leaseId :: T.Text
+  , _renewable :: Bool
+  , _leaseDuration :: Int
+  , _kdata :: a
+  , _wrapInfo :: Maybe ()
+  , _warnings :: Maybe ()
+  , _auth :: Maybe ()
```
Heimdell, line 72: Those three are guaranteed to be null? What if they suddenly aren't?

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,342 @@
...
+-- '/v1/<MOUNT>/metadata/<PATH>'. This is all actually always wrapped
+-- in 'KvResponse' at the 'data' key.
+-- For example:
+--
+-- { "keys":
+--   [ "a1"
+--   , "q1"
+--   ]
+-- }
+newtype ListSecrets = ListSecrets [T.Text]
```
Heimdell, line 89:
```suggestion
newtype ListSecrets = ListSecrets { unListSecrets :: [T.Text] }
```

Again, styleguide.

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,342 @@
...
+  parseJSON = A.withObject "KvResponse" $ \o ->
+    KvResponse
+    <$> o .: "request_id"
+    <*> o .: "lease_id"
+    <*> o .: "renewable"
+    <*> o .: "lease_duration"
+    <*> (o .: "data" >>= A.parseJSON)
+    <*> pure Nothing
+    <*> pure Nothing
+    <*> pure Nothing
```
Heimdell, line 219: Oh

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,342 @@
...
+
+removeNull :: [AT.Pair] -> [AT.Pair]
+removeNull = filter (\(_, v) -> case v of A.Null -> False ; _ -> True)
```
Heimdell, line 222:
```suggestion
removeNull = filter \case
  (_, A.Null) -> False
  _ -> True
```

Heimdell: `BlockArgument` is nice

lib/Entry.hs
Heimdell: (everywhere!)

lib/Entry.hs
```diff
@@ -0,0 +1,115 @@
...
+  , _tags = []
+  }
+
+makeLensesFor [("_path", "path"), ("_masterField", "masterField"), ("_fields", "fields"), ("_tags", "tags")] ''Entry
+
+class DateModified a where
+  dateModified :: Lens' a DateTime
+
+instance DateModified Entry where
+  dateModified = lens _eDateModified (\e d -> e { _eDateModified = d } )
```
Heimdell, line 109:
```hs
makeLenses ''Entry
...
instance DateModified Entry where
  dateModified = eDateModified
```

lib/Entry/Json.hs
```diff
@@ -0,0 +1,84 @@
...
+          [ "date_modified" A..= (field ^. E.dateModified)
+          , "private" A..= (field ^. E.private)
+          , "value" A..= (field ^. E.value)
+          ]
+        from (A.Object o) = do
+          dateModified <- HS.lookup "date_modified" o
+                               >>= \case A.String t -> Just t ; _ -> Nothing
+                               >>= iso8601ParseM . T.unpack
+          value <- HS.lookup "value" o
+                     >>= \case A.String t -> Just t ; _ -> Nothing
```
Heimdell, line 41:
```suggestion
                     >>= \case 
                            A.String t -> Just t
                            _ -> Nothing
```

lib/Entry/Json.hs
```diff
@@ -0,0 +1,84 @@
...
+          _private <- HS.lookup "private" o
+                        >>= \case A.Bool b -> Just b ; _ -> Nothing
+
+          pure $
```
Heimdell, line 45:
```suggestion
          pure
```

Heimdell (commented):
Please abide to Serokell styleguide, which includes indentation, `DerivingStrategies` and other stuff. This is a public project, after all.

lib/Entry/Json.hs
```diff
@@ -0,0 +1,84 @@
...
+import Data.Time.Calendar.OrdinalDate (fromMondayStartWeek)
+import qualified Data.Vector as V
+
+newtype JsonEntry = JsonEntry A.Value
+  deriving (Show, Generic)
+
+instance A.ToJSON JsonEntry
+instance A.FromJSON JsonEntry
+
+fieldConverter :: (Prism A.Value A.Value E.Field E.Field)
```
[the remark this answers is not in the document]
MagicRB: `JsonEntry` is really an example of how to use the interface. The plan is to convert the Vault backend to use it too. And the editor mode on the front end should also use it.

lib/Backend.hs
```diff
@@ -0,0 +1,43 @@
...
+import qualified Toml
+import qualified Data.HashMap.Strict as HS
+
+import Error                         (CofferError)
+import Polysemy.Error                (Error)
+import Toml                          (TomlCodec)
+
+import Polysemy
+
+-- @TODO - rename Secret to Entry?
```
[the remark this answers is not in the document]
MagicRB: yeah, will do, renaming will affect the whole codebase, thats why i waited and didnt rename

lib/Entry.hs
MagicRB: Ok, I didn't know either of those things. I'll try to tackle all of it

lib/Entry.hs
MagicRB: this set? https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Set.html

lib/Backend/Vault/Kv/Internal.hs
MagicRB: They are not, but I couldn't figure out what they are supposed to be, or even could be as Vaults documentation sucks, much of this is reverse engineered... If someone manages to get a response from Vault where those 3 aren't null we can tackle it but right now all responses I've seen had those set to `null`

lib/Entry.hs
MagicRB: I'll do the `Mk` thing later as that will affect all our other branches, let's not make this merge any more confusing and horrible

lib/Backend/Vault/Kv.hs
MagicRB:
```haskell
lib/Vault/Kv.hs:129:16: error:
    • Couldn't match expected type ‘[T.Text]
                                    -> IO (I.KvResponse I.ListSecrets)’
                  with actual type ‘t0’
        ‘t0’ is untouchable
          inside the constraints: x ~ [T.Text]
          bound by a pattern with constructor:
                     ListSecrets :: forall k (m :: k). [T.Text] -> Vault m [T.Text],
                   in a case alternative
          at lib/Vault/Kv.hs:128:7-22
    • In the first argument of ‘embed’, namely
        ‘((perform I.listSecrets) path)’
      In the first argument of ‘(<&>)’, namely
        ‘embed ((perform I.listSecrets) path)’
      In the first argument of ‘(<&>)’, namely
        ‘embed ((perform I.listSecrets) path) <&> (^. I.ddata)’
    • Relevant bindings include
        perform :: Getting
                     (T.Text -> I.VaultToken -> t0)
                     (Vault.Kv.Internal.Routes (Servant.Client.Generic.AsClientT IO))
                     (T.Text -> I.VaultToken -> t0)
                   -> t0
          (bound at lib/Vault/Kv.hs:70:7)
    |
129 |         embed ((perform I.listSecrets) path) <&> (^. I.ddata) <&> \(I.ListSecrets list) -> list
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cabal: Failed to build coffer-0.1.0.0 (which is required by exe:coffer from
coffer-0.1.0.0).
```

I'm getting that error, but I don't get at all why I'm getting it

lib/Backend/Vault/Kv.hs
MagicRB: The name may be bad, it's supposed to convey: "If Nothing, throw this specific error specific to this module" It's just a helper to make the code a tad more readable.

lib/Backend.hs
```diff
@@ -0,0 +1,43 @@
...
+  _codecRead :: Toml.TomlEnv a
+  _codecWrite :: a -> Toml.TomlState a
+  _runEffect :: Member (Embed IO) r
+             => Member (Error CofferError) r
+             => a
+             -> Sem (BackendEffect ': r) t
+             -> Sem r t
+
+data BackendPacked where
+  PackBackend :: Backend a => a -> BackendPacked
```
dcastro, line 40: When you have a existential data type wrapper for a type named `X`, the usual naming convention is to name the wrapper `SomeX`. In this case, it would be `SomeBackend`.

lib/Entry.hs
dcastro: @Heimdell Are there more beside `To/FromJSON`?

lib/Entry.hs
dcastro: > let's not make this merge any more confusing

Agreed. @Heimdell can you please create an issue so we won't forget to do this later?

lib/Entry.hs
dcastro: Yes

lib/Backends.hs
```diff
@@ -0,0 +1,36 @@
...
+  , backendPackedCodec
+  ) where
+
+import qualified Data.Text           as T
+import qualified Toml
+
+import           Backend.Vault.Kv    (VaultKvBackend)
+import           Toml                (TomlCodec)
+import           Backend             (BackendPacked (..), Backend (..))
+import qualified Data.HashMap.Strict as HS
```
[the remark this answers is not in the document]
dcastro: We should have a separate issue for using `stylish-haskell` and applying the conventions across the board. @Heimdell  Can you please create an issue for that?

lib/Backend/Vault/Kv/Internal.hs
dcastro: You mean `LambdaCase` :P

lib/Backend.hs
```diff
@@ -0,0 +1,43 @@
...
+import Toml                          (TomlCodec)
+
+import Polysemy
+
+-- @TODO - rename Secret to Entry?
+data BackendEffect m a where
+  WriteSecret  :: E.Entry -> BackendEffect m Int
+  ReadSecret   :: [T.Text] -> Maybe Int -> BackendEffect m (Maybe E.Entry)
+  ListSecrets  :: [T.Text] -> BackendEffect m (Maybe [T.Text])
+  DeleteSecret :: [T.Text] -> BackendEffect m ()
```
dcastro, line 26: Can you please add haddocks to each operation? In particular, it would be useful to document operational details and invariants that weren't/can't be encoded at the type-level.

For example, we should document that:
1. `WriteSecret` overwrites any entry that might already exist at that path
2. `WriteSecret` does _not_ overwrite directories. If a directory with that path already exists, you'll end up with an entry _and_ a directory sharing the same path.
3. `ListSecret` returns path segments: if the segment is suffixed by `/` then that indicates a directory; otherwise it's an entry.
4. Once all entries are deleted from a directory, then the directory disappears (i.e. `ListSecrets` will no longer list that directory).

There may be more important details I'm missing.

lib/Backend.hs
```diff
@@ -0,0 +1,43 @@
...
+import Error                         (CofferError)
+import Polysemy.Error                (Error)
+import Toml                          (TomlCodec)
+
+import Polysemy
+
+-- @TODO - rename Secret to Entry?
+data BackendEffect m a where
+  WriteSecret  :: E.Entry -> BackendEffect m Int
+  ReadSecret   :: [T.Text] -> Maybe Int -> BackendEffect m (Maybe E.Entry)
```
dcastro, line 24: The `Int` in `WriteSecret` and the `Maybe Int` in `ReadSecret` are version numbers specific to the vault implementation, right?
But we're not handling versioning, and other `BackendEffect` implementations might not support versioning at all. So maybe these two operations should not take/return version numbers?

lib/Backends.hs
```diff
@@ -0,0 +1,36 @@
...
+import Validation (Validation(Failure))
+
+backendPackedCodec
+  :: (T.Text -> Either Toml.TomlBiMapError (Toml.TomlEnv BackendPacked))
+  -> TomlCodec BackendPacked
+backendPackedCodec backends = Toml.Codec input output
```
dcastro, line 18: this function is parameterized over `backends :: (T.Text -> Either Toml.TomlBiMapError (Toml.TomlEnv BackendPacked))`.

But that parameter seems to always be `supportedBackends`. So maybe `backendPackedCodec` should use the `supportedBackends` value directly, and not require that it be passed as a parameter.

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,336 @@
...
+    <*> o .: "renewable"
+    <*> o .: "lease_duration"
+    <*> (o .: "data" >>= A.parseJSON)
+
+removeNull :: [AT.Pair] -> [AT.Pair]
+removeNull = filter (\(_, v) -> case v of A.Null -> False ; _ -> True)
+
+instance A.ToJSON PostSecret where
+  toJSON postSecret =
+    A.object [ "options" A..= A.object (removeNull [ "cas" A..= (postSecret ^. cas)  ])
```
dcastro, line 220: We should just hide the `.=` from `Control.Lens`, and then we can use aeson's `.=` unqualified.

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,336 @@
...
+             , "data" A..= ((postSecret ^. ddata) :: HS.HashMap T.Text T.Text)
```
dcastro, line 221:
```suggestion
             , "data" A..= (postSecret ^. ddata :: HS.HashMap T.Text T.Text)
```

dcastro, line 221: The reason why you had to use a type annotation `:: HashMap Text Text` here is because:

1. The `ddata` lens is polymorphic (`ddata :: Lens' a b`), and
2. GHC knows what `a` is, but has no way of figuring what `b` should be.

This could be resolved by using `FunctionalDependencies`. 

The typeclasses generated by `makeLensesWith` already use `FunctionalDependencies`, so when we migrate to `makeLensesWith`, I'm pretty sure we should be able to delete this type annotation.

dcastro (changes requested):

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+
+        case response ^. I.ddata . at ("version" :: T.Text) of
+          Just (A.Number i) -> maybe (throw MarshallingFailed) pure (S.toBoundedInteger i)
+          _ -> throw MarshallingFailed
+      ReadSecret path version ->
+        runMaybe $ embedCatchMaybe path (readSecret env path version)
+        >>= \(I.KvResponse _ _ _ _ (I.ReadSecret _data _ _ _ _ _) _ _ _) -> do
+          cofferSpecials :: CofferSpecials <-
+            maybeThrow (_data ^.at "#$coffer" >>= A.decodeStrict' . T.encodeUtf8)
+          let secrets = HS.toList $ foldr HS.delete _data ["#$coffer"]
```
dcastro, line 154: No need for `foldr` here, you're folding over a list of one.

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+          let keyToField key = do
+               _modTime <- cofferSpecials ^. fields.at key <&> (^. dateModified)
+               _private <- cofferSpecials ^. fields.at key <&> (^. private)
```
dcastro, line 157: You're mixing lens-style with non-lens-style in the same expression.

This could alternatively be expressed as:
```suggestion
               _private <- cofferSpecials ^? fields . at key . _Just . private
```

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+
+    case x of
+      WriteSecret entry -> do
+        let
+          cofferSpecials = CofferSpecials
+            { _masterKey =
+              entry ^. E.masterField <&> E.getFieldKey
+            , _globalDateModified = entry ^. E.dateModified
+            , _fields =
+              entry ^. E.fields & traverse %~
```
dcastro, line 126: When using `lens`, `each` is more idiomatic than `traverse`. We should do that here (and in a few other places where we're using `traverse` in a lens expression).

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+               _value <- _data ^.at key
+               _key <- E.newFieldKey key
+
+               Just (_key
+                    , E.newField _modTime _value
+                      & E.private .~ _private
+                    )
+
+          fields <- maybeThrow $
+            secrets & traverse %~ (keyToField . fst) & sequence <&> HS.fromList
```
dcastro, line 167: 1. `secrets :: [(Text, Text)]` contains both the keys and the values of all fields
2. The we use `fst` to throw away the value and keep only the key
3. Immediately after that, we call `keyToField` which looks up the field value again (here: `_value <- _data ^.at key`)

This could be avoided. We shouldn't use `fst` to throw away the value.

dcastro (changes requested):

dcastro (changes requested):

lib/Config.hs
```diff
@@ -0,0 +1,41 @@
...
+  }
+  deriving (Show)
+
+
+configCodec :: TomlCodec Config
+configCodec = Config
+  <$> Toml.dimap hsToList listToHs
+  (Toml.list (backendPackedCodec supportedBackends) "backend") Toml..= backends
+  <*> Toml.text "main_backend" Toml..= mainBackend
+  where hsToList hs = snd <$> HS.toList hs
```
dcastro, line 40: `hsToList` is just [`Data.Foldable.toList`](https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-Foldable.html#v:toList)

dcastro (changes requested):

dcastro (changes requested):

lib/Backend/Vault/Kv.hs
Heimdell: > `untouchable`

That means `forall`-s are involved and the function just requires an explicit signature.

lib/Backend/Vault/Kv.hs
Heimdell: In that case, I suggest making it 
```hs
orThrow :: Member (Error e) r => Maybe a -> e -> Sem r a
```
after merge, which will allow using it in various other places.

This order of arguments would make it easy to use it as operator.
```hs
mbSomething `orThrow` MarshallingFailed
```

lib/Backend/Vault/Kv/Internal.hs
Heimdell: _And_ `BlockArguments`, because otherwise it will be `filter $ \case` :D

lib/Backends.hs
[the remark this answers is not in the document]
Heimdell: Will do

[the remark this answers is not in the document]
Heimdell: Dun

coffer.cabal
```diff
@@ -0,0 +1,98 @@
...
+-- A longer description of the package.
+-- description:
+
+-- A URL where users can report bugs.
+-- bug-reports:
+
+-- The license under which the package is released.
+-- license:
+author:             Magic_RB
+maintainer:         magic_rb@redalder.org
```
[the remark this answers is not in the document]
DK318: Done

lib/Entry.hs
```diff
@@ -0,0 +1,108 @@
...
+import qualified Data.HashMap.Strict as HS
+
+import Control.Lens
+import qualified Data.Aeson.Types as A
+import Data.Hashable (Hashable)
+import GHC.Generics (Generic)
+import Data.Time (UTCTime)
+import Language.Haskell.TH.Lens (_CaseE)
+
+type DateTime = UTCTime
```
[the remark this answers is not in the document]

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+
+runVaultIO :: Member (Embed IO) r
+           => Member (Error CofferError) r
+           => BaseUrl
+           -> I.VaultToken
+           -> T.Text
+           -> Sem (BackendEffect ': r) a
+           -> Sem r a
+runVaultIO url token mount = interpret $
+  \x -> do
```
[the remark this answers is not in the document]

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+                { _dateModified = f ^. E.dateModified
+                , _private = f ^. E.private
+                })
+              & HS.mapKeys E.getFieldKey
+            , _tags = map E.getEntryTag $ entry ^. E.tags
+            }
+          secret = I.PostSecret
+            { I._cas = Nothing
+            , I._pdata =
+                  HS.insert "#$coffer" (TL.toStrict . TL.decodeUtf8 $ A.encode cofferSpecials)
```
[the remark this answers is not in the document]

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+          fields <- maybeThrow $
+            secrets & traverse %~ (keyToField . fst) & sequence <&> HS.fromList
+          _tags <- maybeThrow $ cofferSpecials ^. tags & mapM E.newEntryTag
+
+          pure $ E.newEntry path (cofferSpecials ^. globalDateModified)
+            & E.masterField .~ (cofferSpecials ^. masterKey >>= E.newFieldKey)
+            & E.fields .~ fields
+            & E.tags .~ _tags
+      ListSecrets path ->
+        runMaybe (embedCatchMaybe path (listSecrets env path) <&> (^. I.ddata) <&> \(I.ListSecrets list) -> list)
```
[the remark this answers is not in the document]

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+      DeleteSecret path ->
+        embedCatch path (deleteSecret env path) <&> const ()
```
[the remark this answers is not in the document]

lib/Backend/Vault/Kv.hs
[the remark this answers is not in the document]

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+            , I._pdata =
+                  HS.insert "#$coffer" (TL.toStrict . TL.decodeUtf8 $ A.encode cofferSpecials)
+                . HS.map (^. E.value)
+                . HS.mapKeys E.getFieldKey
+                $ entry ^. E.fields
+            }
+
+        response <- embedCatch (entry ^. E.path) (postSecret env (entry ^. E.path) secret)
+
+        case response ^. I.ddata . at ("version" :: T.Text) of
```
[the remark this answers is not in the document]

lib/Backend/Vault/Kv.hs

lib/Backend/Vault/Kv.hs

lib/Backend/Vault/Kv.hs

lib/Backend/Vault/Kv.hs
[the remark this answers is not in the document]

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+                    , E.newField _modTime _value
+                      & E.private .~ _private
+                    )
+
+          fields <- maybeThrow $
+            secrets & traverse %~ (keyToField . fst) & sequence <&> HS.fromList
+          _tags <- maybeThrow $ cofferSpecials ^. tags & mapM E.newEntryTag
+
+          pure $ E.newEntry path (cofferSpecials ^. globalDateModified)
+            & E.masterField .~ (cofferSpecials ^. masterKey >>= E.newFieldKey)
```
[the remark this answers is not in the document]

lib/Backend.hs
DK318: Done

lib/Entry.hs
```diff
@@ -0,0 +1,108 @@
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+module Entry
+  ( dateModified
+  , Entry, EntryConvertible (..), newEntry
+  , path, masterField, fields
+  , Field (..), FieldKey,  newField, getFieldKey
+  , newField, newFieldKey, getFieldKey, newEntryTag, getEntryTag
```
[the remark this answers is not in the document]

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,336 @@
...
+      <*> metadata .: "version"
+
+instance A.FromJSON a => A.FromJSON (KvResponse a) where
+  parseJSON = A.withObject "KvResponse" $ \o ->
+    KvResponse
+    <$> o .: "request_id"
+    <*> o .: "lease_id"
+    <*> o .: "renewable"
+    <*> o .: "lease_duration"
+    <*> (o .: "data" >>= A.parseJSON)
```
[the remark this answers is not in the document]
DK318: Done

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,336 @@
...
+instance A.FromJSON ListSecrets where
+  parseJSON = A.withObject "ListSecrets" $ \o ->
+    ListSecrets <$> o .: "keys"
+
+instance A.FromJSON ReadSecret where
+  parseJSON = A.withObject "ReadSecret" $ \o -> do
+    metadata <- o .: "metadata"
+    ReadSecret
+      <$> o .: "data"
+      <*> (metadata .: "custom_metadata" <|> pure (HS.fromList []))
```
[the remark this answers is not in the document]
DK318: Done

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,336 @@
...
+  ddata :: Lens' a b
+
+instance a ~ b => DataLens (KvResponse a) b where
+  ddata = kdata
+
+instance DataLens PostSecret (HS.HashMap T.Text T.Text) where
+  ddata = pdata
+
+class CustomMetadataLens a where
+  customMetadata :: Lens' a (HS.HashMap T.Text T.Text)
```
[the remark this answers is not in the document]
DK318: Done

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,336 @@
...
+--   { "created_time": "2018-03-22T02:24:06.945319214Z",
+--     "custom_metadata":
+--     { "owner": "jdoe",
+--       "mission_critical": "false"
+--     },
+--     "deletion_time": "",
+--     "destroyed": false,
+--     "version": 2
+--   }
+-- }
```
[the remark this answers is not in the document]
DK318: Done

lib/Backend/Vault/Kv/Internal.hs
DK318: Done

lib/Backend/Vault/Kv/Internal.hs
DK318: Done

DK318: Done

lib/Entry.hs

lib/Backend/Vault/Kv.hs

lib/Backend/Vault/Kv.hs

lib/Backend/Vault/Kv/Internal.hs
DK318: Done

lib/Backend/Vault/Kv/Internal.hs
DK318: Done

lib/Entry.hs

lib/Entry/Json.hs
DK318: Done

lib/Entry/Json.hs
DK318: Done

lib/Entry/Json.hs
```diff
@@ -0,0 +1,84 @@
...
+          dateModified <- HS.lookup "date_modified" o
+                               >>= \case A.String t -> Just t ; _ -> Nothing
+                               >>= iso8601ParseM . T.unpack
+          value <- HS.lookup "value" o
+                     >>= \case A.String t -> Just t ; _ -> Nothing
+          _private <- HS.lookup "private" o
+                        >>= \case A.Bool b -> Just b ; _ -> Nothing
+
+          pure $
+              E.newField dateModified value
```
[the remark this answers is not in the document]
DK318: Done

lib/Entry/Json.hs
```diff
@@ -0,0 +1,84 @@
...
+                dateModified <- HS.lookup "date_modified" o
+                  >>= \case A.String t -> Just t ; _ -> Nothing
+                  >>= iso8601ParseM . T.unpack
+                let _masterField = HS.lookup "master_field" o
+                                     >>= \case A.String t -> E.newFieldKey t ; _ -> Nothing
+                _fields <- HS.lookup "fields" o
+                  >>= (\case A.Object t -> Just t ; _ -> Nothing)
+                  <&> HS.map (^? fieldConverter)
+                  >>= (mapM (uncurry (liftA2 (,)) . over _1 E.newFieldKey) . HS.toList)
+                  <&> HS.fromList
```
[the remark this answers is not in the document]
DK318: Done

lib/Backend/Vault/Kv/Internal.hs
dcastro: > And BlockArguments, because otherwise it will be filter $ \case :D

Oh right, I'm used to associating `BlockArguments` with `do`, but you're absolutely right :D

lib/Backends.hs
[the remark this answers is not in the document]
dcastro: Thank you!

lib/Backend/Vault/Kv.hs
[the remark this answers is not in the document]
dcastro: We don't want to use `error` though - we want to throw a `CofferError`

lib/Backend/Vault/Kv/Internal.hs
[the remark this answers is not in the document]
dcastro:
```
 <|> error "Cannot parse 'custom_metadata' field
```

No need for `<|>` here. Simply using `.:!` and `.!=` will ensure the entire `ReadSecret` parser fails if the `custom_metadata` key has an unexpected format.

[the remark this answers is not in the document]
dcastro: Also, I would use `.:?` instead of `.:!` here.

From the `.:?` docs:

> The result is Nothing if the key is not present or if its value is Null

From the `.:!` docs:

> This differs from .:? by attempting to parse Null the same as any other JSON value, instead of interpreting it as Nothing.

We want `"custom_metadata" : null` to be parsed as `_rsCustomMetadata = Nothing`, so `.:?` seems a better fit.

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,330 @@
...
+removeNull :: [AT.Pair] -> [AT.Pair]
+removeNull = filter \case 
+  (_, Null) -> False 
+  _         -> True
+
+instance ToJSON PostSecret where
+  toJSON postSecret =
+    object [ "options" .= object (removeNull [ "cas" .= (postSecret ^. cas)  ])
+             , "data" .= (postSecret ^. ddata)
+             ]
```
dcastro, line 216: Because we removed the `A.` qualified, the `[` and `,` aren't aligned anymore.

The [styleguide](https://github.com/serokell/style/blob/master/haskell.md#list-declarations) recommends multi-line lists should go in a separate line - to avoid alignment issues during refactoring.

```suggestion
    object
      [ "options" .= object (removeNull [ "cas" .= (postSecret ^. cas)  ])
      , "data" .= (postSecret ^. ddata)
      ]
```

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,330 @@
...
+  { _secret :: HS.HashMap T.Text T.Text
+  , _rsCustomMetadata :: HS.HashMap T.Text T.Text
+  , _createdTime :: T.Text
+  , _deletionTime :: T.Text
+  , _destroyed :: Bool
+  , _version :: Int
+  }
+  deriving (Show)
+makeLenses ''ReadSecret
+makeLensesWith abbreviatedFields ''ReadSecret
```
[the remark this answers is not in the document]
DK318: But we want some lenses to be generated (i.e. all lenses for `UpdateMetadata`)

test/Doctests.hs
```diff
@@ -0,0 +1,7 @@
+module Main where
+
+import Test.DocTest (mainFromCabal)
+import System.Environment (getArgs)
+
+main :: IO ()
+main = mainFromCabal "coffer" =<< getArgs
```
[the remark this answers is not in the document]
DK318: Some doctests are failing because type check failed and I don't know how to fix it =(

lib/Backend/Vault/Kv/Internal.hs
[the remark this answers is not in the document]
DK318: Done

lib/Backend.hs
DK318: Have written some haddoc

lib/Backend.hs
DK318: Done

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+        Just d  -> Toml.codecWrite codec d >>= maybe empty pure . matchA
+    }
+
+
+vaultKvCodec :: TomlCodec VaultKvBackend
+vaultKvCodec = VaultKvBackend
+                  <$> Toml.text "name" Toml..= _vbName
+                  <*> didimatch baseUrlToText textToBaseUrl (Toml.text "address") Toml..= _vbAddress
+                  <*> Toml.text "mount" Toml..= _vbMount
+                  <*> didimatch tokenToText textToToken (Toml.text "token") Toml..= _vbToken
```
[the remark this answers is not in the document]
DK318: I don't get on how to deal with `textToBaseUrl`

lib/Backend.hs
```diff
@@ -0,0 +1,43 @@
...
+  ListSecrets  :: [T.Text] -> BackendEffect m (Maybe [T.Text])
+  DeleteSecret :: [T.Text] -> BackendEffect m ()
+makeSem ''BackendEffect
+
+class Show a => Backend a where
+  _name :: a -> T.Text
+  _codecRead :: Toml.TomlEnv a
+  _codecWrite :: a -> Toml.TomlState a
```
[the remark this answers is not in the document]
DK318: Done

lib/Config.hs
DK318: Done

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+      DeleteSecret path ->
+        embedCatch path (deleteSecret env path) <&> const ()
+  where postSecret env = (I.routes env ^. I.postSecret) mount token
+        readSecret env = (I.routes env ^. I.readSecret) mount token
+        listSecrets env = (I.routes env ^. I.listSecrets) mount token
+        updateMetadata env = (I.routes env ^. I.updateMetadata) mount token
+        deleteSecret env = (I.routes env ^. I.deleteSecret) mount token
+
+        exceptionHandler :: [T.Text] -> ClientError -> Maybe CofferError
+        exceptionHandler path =
```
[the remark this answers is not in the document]

lib/Backend/Vault/Kv/Internal.hs
DK318: Done

lib/Backend/Vault/Kv/Internal.hs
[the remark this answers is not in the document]
dcastro: Yes, and `makeLensesWith` will generate those lenses (see my example [here](https://github.com/serokell/coffer/pull/7#discussion_r814922853)).

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,238 @@
...
+            , I._psDdata =
+                    HS.insert "#$coffer" (TL.toStrict $ A.encodeToLazyText cofferSpecials)
+                  . HS.map (^. E.value)
+                  . HS.mapKeys E.getFieldKey
+                  $ entry ^. E.fields
+              }
+
+        response <- embedCatch (entry ^. E.path) (postSecret env (entry ^. E.path) secret)
+
+        case response ^. I.ddata . at "version" of
```
dcastro, line 147: Now that `WriteSecret` no longer return the version number, there's no need to deserialize the response.

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -252,29 +252,30 @@ type VaultTokenHeader = Header' '[Strict, Required] "X-Vault-Token" VaultToken
...
   { -- | To read a secret under a path use `readSecret`
-    -- >>> (routes clientEnv ^. readSecret) vaultToken "kv" ["testbed", "a1"]
+    --  (routes clientEnv ^. readSecret) vaultToken "kv" ["testbed", "a1"]
     _readSecret :: route
     :- "v1"
     :> Capture "mount" T.Text
     :> "data"
     :> VaultTokenHeader
     :> CaptureAll "segments" T.Text
-    :> QueryParam "version" Int
```
dcastro: We shouldn't remove `version` from `Routes._readSecret`. These routes are supposed to match Vault's HTTP API, so it's OK to have a `version` query parameter here.

In my other comment, I meant we should remove versions from our `BackendEffect` abstraction only.

lib/Backends.hs
```diff
@@ -0,0 +1,36 @@
...
+                         case Toml.backward Toml._Text t >>= backends of
+                           Right c -> c toml
+                           Left e -> Failure
+                                     [ Toml.BiMapError "type" e
+                                     ]
+                       Nothing -> Failure
+                                  [ Toml.BiMapError "type" $ Toml.ArbitraryError
+                                    "Backend doesn't have a `type` key"
+                                  ]
+        output (PackBackend a) =  PackBackend <$> _codecWrite a
```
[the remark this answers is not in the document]
dcastro: Did you forget to push something? Doesn't look like this has been fixed.

lib/Backend.hs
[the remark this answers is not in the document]
dcastro: Did you forget to push something? Doesn't look like this has been fixed.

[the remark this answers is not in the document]
dcastro: Did you forget to push something? Doesn't look like this has been fixed.

lib/Backend/Vault/Kv.hs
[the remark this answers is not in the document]
dcastro: Line 73 should continue using `didimatch`, I was only referring to line 75.

lib/Backends.hs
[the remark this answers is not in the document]
DK318: Fixed

lib/Backend.hs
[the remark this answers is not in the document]
DK318: Fixed

lib/Backend/Vault/Kv/Internal.hs
DK318: Then I don't understand how to remove `version` in `ReadSecret`. We have this code
```Haskell
...
ReadSecret path ->
        embedCatchMaybe path (readSecret env path) >>= \case
...
readSecret env = (I.routes env ^. I.readSecret) mount token
...
```
And we need the `version` parameter in `readSecret`.

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,336 @@
...
+-- , "custom_metadata":
+--   { "example": ":)"
+--   }
+-- }
+data UpdateMetadata =
+  UpdateMetadata
+  { _maxVersions :: Maybe Int
+  , _casRequired :: Maybe Bool
+  , _deleteVersionAfter :: Maybe T.Text
+  , _uCustomMetadata :: HS.HashMap T.Text T.Text
```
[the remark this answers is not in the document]
DK318: Done

lib/Backend/Vault/Kv.hs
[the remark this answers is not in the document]

lib/Config.hs
DK318: Fixed

lib/Backend/Vault/Kv/Internal.hs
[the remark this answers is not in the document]
DK318: Done

lib/Backend.hs
```diff
@@ -0,0 +1,47 @@
...
+  _runEffect :: Member (Embed IO) r
+             => Member (Error CofferError) r
+             => a
+             -> Sem (BackendEffect ': r) t
+             -> Sem r t
+
+data BackendPacked where
+  SomeBackend :: Backend a => a -> BackendPacked
```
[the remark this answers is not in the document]
DK318: Done

lib/Backend.hs
```diff
@@ -0,0 +1,47 @@
...
+
+import Error                         (CofferError)
+import Polysemy.Error                (Error)
+import Toml                          (TomlCodec)
+
+import Polysemy
+
+-- @TODO - rename Secret to Entry?
+data BackendEffect m a where
+  WriteSecret  :: E.Entry -> BackendEffect m () -- ^ overwrites any entry that might already exist at that path.
```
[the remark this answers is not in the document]
DK318: Done

lib/Backend/Vault/Kv.hs

lib/Backend/Vault/Kv/Internal.hs
dcastro: `readSecret`'s version parameter is optional, we can pass it `Nothing`:

```hs
ReadSecret path ->
        embedCatchMaybe path (readSecret env path Nothing) >>= \case
```

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,310 @@
...
+      ]
+
+instance ToJSON UpdateMetadata where
+  toJSON updateMetadata =
+    object (removeNull 
+              [ "max_versions" .= (updateMetadata ^. maxVersions)
+              , "cas_required" .= (updateMetadata ^. casRequired)
+              , "delete_version_after" .= (updateMetadata ^. deleteVersionAfter)
+              , "custom_metadata" .= (updateMetadata ^. customMetadata)
+              ])
```
dcastro, line 223: Let's remove excessive indentation here as well.

```suggestion
    object $ removeNull 
      [ "max_versions" .= (updateMetadata ^. maxVersions)
      , "cas_required" .= (updateMetadata ^. casRequired)
      , "delete_version_after" .= (updateMetadata ^. deleteVersionAfter)
      , "custom_metadata" .= (updateMetadata ^. customMetadata)
      ]
```

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,242 @@
...
+        embedCatch path (void $ deleteSecret env path)
+
+  where 
+    postSecret env = (I.routes env ^. I.postSecret) mount token
+    readSecret env = (I.routes env ^. I.readSecret) mount token
+    listSecrets env = (I.routes env ^. I.listSecrets) mount token
+    updateMetadata env = (I.routes env ^. I.updateMetadata) mount token
+    deleteSecret env = (I.routes env ^. I.deleteSecret) mount token
+
+    -- | Handles @ClientError@ in the next way:
```
dcastro, line 195:
```suggestion
    -- | Handles @ClientError@ in the following way: 
```

lib/Backend/Vault/Kv/Internal.hs
DK318: Done

lib/Backend/Vault/Kv/Internal.hs
```diff
@@ -0,0 +1,311 @@
...
+-- >    { "example": ":)"
+-- >    }
+-- >  }
+data UpdateMetadata =
+  UpdateMetadata
+  { _umMaxVersions :: Maybe Int
+  , _umCasRequired :: Maybe Bool
+  , _umDeleteVersionAfter :: Maybe T.Text
+  , _umCustomMetadata :: HS.HashMap T.Text T.Text
+  }
```
dcastro, line 168: There's no need for the `_` prefix anymore.

When using `makeLenses`, it makes sense to add a `_` prefix.
`makeLenses` will detect a field named `_x` and create a lens named `x`. See the example in the [docs here](https://hackage.haskell.org/package/lens-5.1/docs/Control-Lens-Combinators.html#v:makeLenses),

But because we're now using `makeLensesWith abbreviatedFields` with an `um` prefix, there's no need for an _additional_  `_` prefix.

lib/Backend/Vault/Kv.hs
```diff
@@ -0,0 +1,231 @@
...
+                      => Maybe a
+                      -> Sem r a
+        maybeThrow = maybe (throw MarshallingFailed) pure
+
+instance Backend VaultKvBackend where
+  _name s = _vbName s
+  _codecRead = Toml.codecRead vaultKvCodec
+  _codecWrite a = Toml.codecWrite vaultKvCodec a
+                  <* Toml.codecWrite (Toml.text "type") "vault"
+  _runEffect b = runVaultIO (_vbAddress b) (_vbToken b) (_vbMount b)
```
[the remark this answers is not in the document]
dcastro: Don't forget about this comment please ^^

lib/Config.hs
```diff
@@ -0,0 +1,42 @@
...
+  deriving (Show)
+
+
+configCodec :: TomlCodec Config
+configCodec = Config
+  <$> Toml.dimap hsToList listToHs
+  (Toml.list backendPackedCodec "backend") Toml..= backends
+  <*> Toml.text "main_backend" Toml..= mainBackend
+  where hsToList = toList
+        listToHs list = HS.fromList $ fmap (\y@(SomeBackend x) -> (_name x, y)) list
```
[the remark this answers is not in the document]
DK318: I've left this for symmetrical naming

lib/Entry/Json.hs
Heimdell: Well, it doesn't look so in the diff :D

lib/Entry/Json.hs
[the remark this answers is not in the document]
Heimdell: Neither does this.

Heimdell (commented):
Looks fine for me. Please fix remaining @dcastro's issues.

lib/Backend/Vault/Kv/Internal.hs
DK318: Done

lib/Config.hs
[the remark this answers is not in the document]
DK318: Done

lib/Backend/Vault/Kv/Internal.hs
DK318: Done

dcastro: Please do the same (use an abbreviated prefix + switch to `makeLensesWith`) for `FieldMetadata`, `CofferSpecials`, `KvResponse` and `Routes`.

dcastro: Oh and remove the `_` prefix from `VaultKvBackend` please.

lib/Backend/Vault/Kv.hs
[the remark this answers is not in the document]
DK318: Done (I think)

lib/Backend/Vault/Kv/Internal.hs
DK318: Done

dcastro (approved):

PR #8, opened 2022-02-21
Supported `view` command by default

### LINKED ISSUE #3: Make the `view` command the default command
There's a good chance `view` will be the command users use the most.

It would be convenient to not have to specify the command name, and just default to using `view` when no command name is given.

In other words, these two should behave the same:

```
$ coffer view path/to/entry
$ coffer path/to/entry
```

DK318:
Resolves #3


MagicRB:
The commit should have be something akin to 
```
[#3] Support view command by default

Problem:


Solution:

```

the problem/solution part may be  unnecessary as it's such a simple change and the problem/solution texts on these commits generally end up being `X is missing.` `Add X.`

dcastro:
> The commit should be something akin to

Absolutely.

> the problem/solution part may be unnecessary as it's such a simple change and the problem/solution texts on these commits generally end up being X is missing. Add X.

Nevertheless, it's still good practice to get in the habit of doing this.

Keep in mind that the "problem" section isn't necessarily just the fact that x is missing. It should also explain why x is needed in the first place. Why are we making `view` the default command? Why is that necessary? What root problem does that actually solve for the user (or someone else)? Every commit should have a clear purpose.

Sometimes this is already described in the corresponding github issue, so using that as a reference for your "Problem" section is a good starting point.

Issue #9, opened 2022-02-21
`create` command allows to have entries and directories with same name


DK318:
E.g. we can do
```
coffer create a/b
coffer create a/b/c
```
Both commands would create entries `a/b` and `a/b/c` which is not good.

PR #10, opened 2022-02-21
[#9] Directory check in `create` command

DK318:
Problem: we can create entries and directories with the same names.
Solution: added subfunction `checkForEntriesInEntryPath` which tries to find an existing entry subpath in given path. Then checked if the given entry path is an existing directory or contains existing entries in subpaths.


lib/Backend/Commands.hs
```diff
@@ -87,7 +88,27 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
       & E.fields .~ HashMap.fromList allFields
       & E.tags .~ tags
 
-  -- TODO: check that a directory does not exist at the given path.
+    checkForEntriesInEntryPath :: Member BackendEffect r => EntryPath -> Sem r (Maybe EntryPath)
+    checkForEntriesInEntryPath entryPath =
+      let 
+        Path.EntryPath (start NE.:| rest) = entryPath
```
dcastro, line 94: I think it's fine to use type/constructor names like `EntryPath` and `Path` unqualified. Same for the `:|` operator.

lib/Backend/Commands.hs
```diff
@@ -87,7 +88,27 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
       & E.fields .~ HashMap.fromList allFields
       & E.tags .~ tags
 
-  -- TODO: check that a directory does not exist at the given path.
+    checkForEntriesInEntryPath :: Member BackendEffect r => EntryPath -> Sem r (Maybe EntryPath)
```
dcastro, line 91: Since the `r` type variable in `createCmd` and in `checkForEntriesInEntryPath` are supposed to unify, you could do this:

1. Add a `forall r.` to `createCmd`
2. In `checkForEntriesInEntryPath`, remove the `Member BackendEffect r => ` constraint.

Thanks to [`ScopedTypeVariables`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/scoped_type_variables.html), GHC will know the `r` in both functions is the same, so any constraints applied to `r` in `createCmd` will also be observed in `checkForEntriesInEntryPath`.

This way, we avoid repeating constraints.

lib/Backend/Commands.hs
```diff
@@ -87,7 +88,27 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
...
+        Path.EntryPath (start NE.:| rest) = entryPath
+        startPath = Path.EntryPath (start NE.:| [])
+        restPath = Path.Path rest
+      in process startPath restPath
+        where 
+          process :: Member BackendEffect r => EntryPath -> Path -> Sem r (Maybe EntryPath)
+          process _ (Path.Path []) = pure Nothing
+          process entryPath (Path.Path (x : xs)) = 
+            pathIsEntry entryPath >>= \case
+              True -> (pure . Just) entryPath
```
dcastro, line 103: Instead of returning the entry here, and then using `when isJust + throw` in line 110, we could simplify this by throwing the error here.

lib/Backend/Commands.hs
```diff
@@ -87,7 +88,27 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
       & E.fields .~ HashMap.fromList allFields
       & E.tags .~ tags
 
-  -- TODO: check that a directory does not exist at the given path.
+    checkForEntriesInEntryPath :: Member BackendEffect r => EntryPath -> Sem r (Maybe EntryPath)
+    checkForEntriesInEntryPath entryPath =
+      let 
+        Path.EntryPath (start NE.:| rest) = entryPath
+        startPath = Path.EntryPath (start NE.:| [])
+        restPath = Path.Path rest
```
dcastro, line 96: We could simplify this by separating these two bits of logic: 1. build a list of all the paths we need to check and then 2. iterate through them and check if they already point to an entry.

I think we could add a function like this to the `Path` module (which might even be useful in other situations):

```hs
entryPathParentDirs :: EntryPath -> NonEmpty Path
```

such that (in pseudocode):

```hs
entryPathParentDirs "/a/b/c" == [  "/", "/a", "/a/b" ]
```

Then, here in `checkForEntriesInEntryPath`, all we had to do was iterate through those paths.

dcastro:
Good work! Just needs a bit of polishing and we're good to go.

Also great work in identifying yet another issue: the fact that we need to check _all_ parent dirs, not just the target dir! :clap: 

Thanks for tagging the PR's title and the commit messages with the issue number `[#x]`

app/Main.hs
```diff
@@ -25,7 +25,7 @@ url :: BaseUrl
 url = BaseUrl Http "127.0.0.1" 8200 ""
 
 token :: VaultToken
-token = VaultToken "s.AnHJuzOxRiD12eLxALFytccC"
```
MagicRB, line 28: Same thing here as in the other PR #13.

Issue #12, opened 2022-02-22
`rename` command


DK318:
`rename` command is not implemented.

PR #13, opened 2022-02-22
[#12] Rename command implemented

DK318:
Problem: `rename` command is not implemented.
Solution: implemented `rename` command via `copy` and `delete` commands.
Fixes: #12

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -114,8 +114,24 @@ main = do
 
           SomeCommand cmd@CmdRename{} -> do
             runCommand cmd >>= \case
-              -- TODO: print result
-              _ -> undefined
+              CPRSuccess copiedPaths ->
+                forM_ copiedPaths \(from, to) ->
+                  printSuccess $ "Renamed '" +| from |+ "' to '" +| to |+ "'."
+              CPRPathNotFound path -> pathNotFound path
+              CPRMissingEntryName -> printError
+                "The destination path is not a valid entry path. Please specify the new name of the entry."
+              CPRDestinationIsDirectory paths -> do
+                let header = "The following entries cannot be renamed because a directory already exists at the destination."
+                let errorMsgs = NE.toList paths <&> \(from, to) -> "Cannot rename '" +| from |+ "' to '" +| to |+ "'."
+                printError $ unlinesF @_ @Builder $ header : "" : errorMsgs
+              CPREntryAlreadyExists paths -> do
+                let header = unlinesF @_ @Builder
+                      [ "The following entries cannot be renamed because an entry already exists at the destination."
+                      , "Use '--force' or '-f' to overwrite existing entries."
+                      ]
+                let errorMsgs = NE.toList paths <&> \(from, to) -> "Cannot rename '" +| from |+ "' to '" +| to |+ "'."
+                printError $ unlinesF @_ @Builder $ header : "" : errorMsgs
+
           SomeCommand cmd@CmdCopy{} -> do
             runCommand cmd >>= \case
               CPRSuccess copiedPaths ->
--- a/lib/Backend/Commands.hs
+++ b/lib/Backend/Commands.hs
@@ -26,7 +26,7 @@ import Data.Set (Set)
 import qualified Data.Set as Set
 
 import Backend (BackendEffect, listSecrets, readSecret, writeSecret, deleteSecret)
-import Entry (Entry, Field, FieldKey, value, fields, dateModified, newEntry, newField, visibility, FieldVisibility(..), EntryTag)
+import Entry (Entry, Field, FieldKey, value, fields, dateModified, newEntry, newField, visibility, FieldVisibility(..), EntryTag, path)
 import qualified Entry as E
 import Coffer.Directory (Directory)
 import qualified Coffer.Directory as Dir
@@ -45,7 +45,7 @@ runCommand = \case
   CmdSetField opts -> catchAndReturn $ setFieldCmd opts
   CmdDeleteField opts -> deleteFieldCmd opts
   CmdFind opts -> findCmd opts
-  CmdRename opts -> renameCmd opts
+  CmdRename opts -> catchAndReturn $ renameCmd opts
   CmdCopy opts -> catchAndReturn $ copyCmd opts
   CmdDelete opts -> catchAndReturn $ deleteCmd opts
   CmdTag opts -> catchAndReturn $ tagCmd opts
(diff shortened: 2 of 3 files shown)


MagicRB:
IMO it should be moved to the backend as it can be done more efficiently for some backends, namely `pass` or if we eventually implement our own secret engine in vault.

dcastro:
> IMO it should be moved to the backend as it can be done more efficiently for some backends, namely `pass` or if we eventually implement our own secret engine in vault.

TBH it's not just this bit, I think a lot of things will be moved to the backend (and some things might be moved in the opposite direction, who knows). But we'll only know for certain when we actually implement those other engines; then we'll have to figure out what they have in common and what they don't, and where the right level of abstraction is.

I think we can leave it here _for now_ and re-evaluate this when it becomes a problem.

app/Main.hs
```diff
@@ -25,7 +25,7 @@ url :: BaseUrl
 url = BaseUrl Http "127.0.0.1" 8200 ""
 
 token :: VaultToken
-token = VaultToken "s.AnHJuzOxRiD12eLxALFytccC"
```
MagicRB, line 28: This will just introduce fuss into the commit history the token should be redacted anyway (not that it's valid... hopefully) but it might give some people false "hope" and might trigger those "secret finding bots" that apparently exist.

MagicRB (changes requested):

lib/Backend/Commands.hs
```diff
@@ -237,10 +233,24 @@ findCmd (FindOptions pathMb textMb sortMb filters filterFields) = do
...
+     , Member (Embed IO) r
+     , Member (Error CofferError) r
+     , Member (Error RenameResult) r
+     ) 
+  => RenameOptions -> Sem r RenameResult
+renameCmd (RenameOptions oldPath newPath force) =
+  copyCmd (CopyOptions oldPath newPath force) >>= \case
+     cSuccess@(CPRSuccess _) -> 
+       if oldPath == newPath then
+         -- if oldPath equals to newPath then, we don't want to accidentally delete renamed directory/entry.
```
dcastro, line 247: Good catch!

dcastro: In fact, the `copyCmd` command should error out when `oldPath == newPath`. That's what the `cp` command does, and it seems like a sensible approach.

```
$ touch a
$ cp a a
cp: 'a' and 'a' are the same file
```

Could you please create a new issue for  this?

DK318: `coffer copy` throws an error if two paths are equal. But if we add `-f` then there would be no errors. Is it expected behavior?

dcastro: > Is it expected behavior

It is the "expected behavior" in the sense that the [specification](https://www.notion.so/serokell/Password-sharing-2-232733503bb4469d85713d08f3b3a5e4#2459058a49df42ed84bcc37dd1d6fd4c) doesn't say anything about this particular scenario (when the 2 paths are the same). So in that sense, `coffer copy` behaving the exact same way for 2 different paths as it does for 2 equal paths is the expected behavior.

But what I'm proposing is that we consider this edge case, and consider changing `coffer copy`'s behavior. I think it should always throw an error when the 2 paths are the same, like the `cp` command, regardless of whether `-f` was used. 

We should create an issue to discuss this and see if the rest of the team agrees.

app/Main.hs

tests/golden/rename-command/rename-command.bats
```diff
@@ -0,0 +1,67 @@
...
+  run coffer create /a/b/c
+  assert_output "[SUCCESS] Entry created at '/a/b/c'."
+
+  run coffer create /d/c
+  assert_output "[SUCCESS] Entry created at '/d/c'."
+
+  run coffer rename /a/b /d
+
+  assert_failure
+  cat rename-without-force.gold | assert_output --partial -
```
dcastro, line 39: The same comment [I made here](https://github.com/serokell/coffer/pull/32#discussion_r820249222) also applies to this test

tests/golden/rename-command/rename-command.bats
```diff
@@ -0,0 +1,67 @@
...
+  assert_output "[SUCCESS] Entry created at '/a/b/c'."
+
+  run coffer create /d/c
+  assert_output "[SUCCESS] Entry created at '/d/c'."
+
+  run coffer rename /a/b /d -f
+  assert_output "[SUCCESS] Renamed '/a/b/c' to '/d/c'."
+
+  run cleanOutput coffer view /
+  cat rename-with-force.gold | assert_output -
```
dcastro, line 53: Checking that `view` outputs a single `/d/c` entry doesn't necessarily prove that `/a/b/c` has been renamed to `/d/c/`.

If there was a bug in `coffer rename` that simply deleted the `/a/b/c` entry without copying it, the test would still pass.

We should make this test more robust. I suggest we add a field named `x` to `/a/b/c`, and a field named `y` to `/d/c`. Then, at the end, we check that `view` returns an entry at `/d/c` _with a field named `x`_.

dcastro (changes requested):

lib/Backend/Commands.hs
```diff
@@ -241,10 +241,24 @@ findCmd (FindOptions pathMb textMb sortMb filters filterFields) = do
...
+         pure cSuccess
+       else do
+         -- If directory/entry was successfully copied, then we can delete old directory/entry without delete errors.
+         void $ catchAndReturn $ deleteCmd (DeleteOptions oldPath True)
```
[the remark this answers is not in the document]

tests/golden/rename-command/rename-command.bats
DK318: Done

tests/golden/rename-command/rename-command.bats
DK318: Done

dcastro (changes requested):

app/Main.hs
```diff
@@ -100,8 +100,23 @@ main = do
```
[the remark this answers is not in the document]
DK318: Changing the commit message would be enough?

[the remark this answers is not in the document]
dcastro: No, because the commit would still contain changes that will need to be squashed against two different commits... 

It needs to be split. You can do this:

* `git reset --soft HEAD~1` to undo the latest commit and keep the changes in the working directory.
* Then stage related to the `Command` module, and run `git commit --fixup 15eff3`
* Then stage related to the tests, and run `git commit --fixup ade525`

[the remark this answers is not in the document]
DK318: Done

tests/golden/rename-command/rename-command.bats
```diff
@@ -0,0 +1,145 @@
...
+  coffer create /a/b/d
+  coffer create /a/b/e
+
+  coffer create /d/c
+  coffer create /d/d
+
+  run coffer rename /a/b /d
+
+  assert_failure
+  assert_output --partial - <<EOF
```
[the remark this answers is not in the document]
DK318: Done

lib/Backend/Commands.hs
```diff
@@ -241,10 +241,44 @@ findCmd (FindOptions pathMb textMb sortMb filters filterFields) = do
...
+renameCmd
+  :: forall r.
+     ( Member BackendEffect r
+     , Member (Embed IO) r
+     , Member (Error CofferError) r
+     , Member (Error RenameResult) r
+     )
+  => RenameOptions -> Sem r RenameResult
+renameCmd (RenameOptions oldPath newPath force) =
+  copyCmd (CopyOptions oldPath newPath False) `catch` handler >>= \case
```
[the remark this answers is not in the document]
DK318: Removed catching

lib/Backend/Commands.hs
```diff
@@ -241,10 +241,44 @@ findCmd (FindOptions pathMb textMb sortMb filters filterFields) = do
...
+     cError -> throw cError
+  where
+    handler :: RenameResult -> Sem r RenameResult
+    handler = \case
+      cClashed@(CPREntryAlreadyExists clashed) ->
+       if force then
+         copyCmd (CopyOptions oldPath newPath True) >>= \case
+          cSuccess@(CPRSuccess copiedPaths) -> do
+            -- we don't want to delete clashed entries if renaming is forced.
+            let filteredCopiedPaths = [(from, to) | (from, to) <- copiedPaths, (_, to_) <- NE.toList clashed, to_ /= from]
```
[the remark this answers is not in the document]

Heimdell (approved):

lib/Backend/Commands.hs
```diff
@@ -241,25 +241,45 @@ findCmd (FindOptions pathMb textMb sortMb filters filterFields) = do
...
+  :: forall r.
+     ( Member BackendEffect r
+     , Member (Embed IO) r
+     , Member (Error CofferError) r
+     , Member (Error RenameResult) r
+     )
+  => RenameOptions -> Sem r RenameResult
+renameCmd (RenameOptions oldPath newPath force) = do
+  operations <- buildCopyOperations oldPath newPath force
+  void $ runCopyOperations operations
```
dcastro, line 254: Hmm it bothers me that we're using `void` to _swallow_ the result of `runCopyOperations` here...

* The type of `runCopyOperations` says it returns a `CopyResult` - which may be a successful `CopyResult` or a failed `CopyResult`. And if it's a failure, we shouldn't ignore it with `void`.
* However, in practice, the implementation of `runCopyOperations` always returns a successful `CopyResult`.

The crux of the problem seems to be that `runCopyOperations'`s type is misleading. It says it may fail or succeed, but that's not really true. Since it never returns a failed `CopyResult`, perhaps we should change its type to return a `Sem r ()` instead.

lib/Backend/Commands.hs
```diff
@@ -241,25 +241,45 @@ findCmd (FindOptions pathMb textMb sortMb filters filterFields) = do
...
+  -- we don't want to delete clashed entries if renaming is forced.
+  let pathsToDelete =
+        flip filter operations \(CopyOperation old _) -> none (\(CopyOperation _ new) -> old ^. path == new ^. path) operations
```
dcastro, line 257: This line is way too long (see [the style guide here](https://github.com/serokell/coffer/blob/main/docs/code_style.md#line-length))

dcastro (changes requested):

lib/Backend/Commands.hs

lib/Backend/Commands.hs

dcastro (approved):

MagicRB (approved):
LGTM

PR #14, opened 2022-02-22
[#11] Checking for entries in destination subpaths in `copy` command

### LINKED ISSUE #11: `copy` command allows creating directory and entry with same names
```
coffer create a/b
coffer create c/d/e
coffer copy c/d/e a/b/c
coffer find
```
This set of commands prints the next thing
```
/
  a/
    b - [2022-02-22 10:43:42]
    b/
      c - [2022-02-22 10:43:57]
  c/
    d/
      e - [2022-02-22 10:43:47]
```
which looks strange

DK318:
Problem: there were no checks for entries in destination subpaths in `copy` command.
Solution: created subfunction `checkForEntriesInPath`, which validates that there are no entries in subpaths of a given path.

Fixes #11

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -148,6 +148,8 @@ main = do
                 let header = "The following entries cannot be copied because a directory already exists at the destination."
                 let errorMsgs = NE.toList paths <&> \(from, to) -> "Cannot copy '" +| from |+ "' to '" +| to |+ "'."
                 printError $ unlinesF @_ @Builder $ header : "" : errorMsgs
+              CPRDestinationHasEntry entry -> printError $
+                "The following entries cannot be copied because a destination directory has an entry '" +| build entry |+ "' in it's subpath."
               CPREntryAlreadyExists paths -> do
                 let header = unlinesF @_ @Builder
                       [ "The following entries cannot be copied because an entry already exists at the destination."
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -5,7 +5,7 @@
 module Main where
 
 import Control.Lens
-import Control.Monad (forM_, when)
+import Control.Monad (forM_, forM, when)
 import Options.Applicative ( execParser )
 import Polysemy
 import Polysemy.Error (errorToIOFinal, Error)
@@ -14,7 +14,6 @@ import Fmt
 import qualified Data.Text.IO as TIO
 import qualified Toml
 import qualified Data.HashMap.Strict as HS
-import qualified Data.List.NonEmpty as NE
 
 import CLI.Parser
 import CLI.Types
@@ -26,6 +25,7 @@ import Coffer.Path ( Path, EntryPath )
 import qualified Entry as E
 import qualified Coffer.Directory as Dir
 import Config (configCodec, Config (..))
+import Entry (path, Entry)
 
 runBackend
   :: SomeBackend
@@ -76,12 +76,9 @@ main = do
           SomeCommand cmd@(CmdCreate opts) -> do
             runCommand cmd >>= \case
               CRSuccess _ -> printSuccess $ "Entry created at '"  +| coPath opts |+ "'."
-              CREntryAlreadyExists path -> printError $ unlinesF
-                [ "An entry already exists at '" <> build path <> "'."
-                , "Use '--force' or '-f' to overwrite it."
-                ]
-              CRDirectoryAlreadyExists path -> printError $ "A directory already exists at '" <> build path <> "'."
-              CRParentPathContainsEntry path -> printError $ "An entry already exists at '" <> build path <> "'."
+              CRCreateError error -> do
+                let errorMsg = createErrorToBuilder error
+                printError $ unlinesF @_ @Builder $ "The entry cannot be created:" : "" : [errorMsg]
 
           SomeCommand cmd@(CmdSetField opts) -> do
             let fieldName = sfoFieldName opts
@@ -122,17 +119,9 @@ main = do
               CPRPathNotFound path -> pathNotFound path
               CPRMissingEntryName -> printError
                 "The destination path is not a valid entry path. Please specify the new name of the entry."
-              CPRDestinationIsDirectory paths -> do
-                let header = "The following entries cannot be renamed because a directory already exists at the destination."
-                let errorMsgs = NE.toList paths <&> \(from, to) -> "Cannot rename '" +| from |+ "' to '" +| to |+ "'."
-                printError $ unlinesF @_ @Builder $ header : "" : errorMsgs
-              CPREntryAlreadyExists paths -> do
-                let header = unlinesF @_ @Builder
-                      [ "The following entries cannot be renamed because an entry already exists at the destination."
-                      , "Use '--force' or '-f' to overwrite existing entries."
-                      ]
-                let errorMsgs = NE.toList paths <&> \(from, to) -> "Cannot rename '" +| from |+ "' to '" +| to |+ "'."
-                printError $ unlinesF @_ @Builder $ header : "" : errorMsgs
+              CPRCreateErrors errors -> do
+                errorMsgs <- buildErrorMessages errors
+                printError $ unlinesF @_ @Builder $ "The following entries cannot be renamed:" : "" : errorMsgs
 
           SomeCommand cmd@(CmdCopy opts) -> do
             runCommand cmd >>= \case
@@ -144,19 +133,9 @@ main = do
               CPRPathNotFound path -> pathNotFound path
               CPRMissingEntryName -> printError
                 "The destination path is not a valid entry path. Please specify the new name of the entry."
-              CPRDestinationIsDirectory paths -> do
-                let header = "The following entries cannot be copied because a directory already exists at the destination."
-                let errorMsgs = NE.toList paths <&> \(from, to) -> "Cannot copy '" +| from |+ "' to '" +| to |+ "'."
-                printError $ unlinesF @_ @Builder $ header : "" : errorMsgs
-              CPRDestinationHasEntry entry -> printError $
-                "The following entries cannot be copied because a destination directory has an entry '" +| build entry |+ "' in it's subpath."
-              CPREntryAlreadyExists paths -> do
-                let header = unlinesF @_ @Builder
-                      [ "The following entries cannot be copied because an entry already exists at the destination."
-                      , "Use '--force' or '-f' to overwrite existing entries."
-                      ]
-                let errorMsgs = NE.toList paths <&> \(from, to) -> "Cannot copy '" +| from |+ "' to '" +| to |+ "'."
-                printError $ unlinesF @_ @Builder $ header : "" : errorMsgs
+              CPRCreateErrors errors -> do
+                errorMsgs <- buildErrorMessages errors
+                printError $ unlinesF @_ @Builder $ "The following entries cannot be copied:" : "" : errorMsgs
 
           SomeCommand cmd@(CmdDelete opts) -> do
             runCommand cmd >>= \case
@@ -200,3 +179,27 @@ main = do
 
       pathNotFound :: Member (Embed IO) r => Path -> Sem r ()
       pathNotFound path = printError $ "Entry or directory not found at '" +| path |+ "'."
+
+      createErrorToBuilder :: CreateError -> Builder
+      createErrorToBuilder = \case
+        CEEntryAlreadyExists entryTo -> unlinesF @_ @Builder
+          [ "An entry already exists at '" +| entryTo ^. path |+ "'."
+          , "Use '--force' or '-f' to overwrite existing entries."
+          ]
+        CEDestinationIsDirectory entryTo -> "'" +| entryTo ^. path |+ "' is a directory."
+        CEParentDirectoryIsEntry (_, clashed) ->
+          "Attempted to create the directory '" +| clashed |+ "' but an entry exists at that path."
+
+      getEntryFromCreateError :: CreateError -> Entry
+      getEntryFromCreateError = \case
+        CEParentDirectoryIsEntry (entryTo, _) -> entryTo
+        CEDestinationIsDirectory entryTo -> entryTo
+        CEEntryAlreadyExists entryTo -> entryTo
+
+      buildErrorMessages :: [(EntryPath, CreateError)] -> Sem r [Builder]
+      buildErrorMessages errors = do
+        forM errors \(from, err) -> do
+          let entryTo = getEntryFromCreateError err
+          let header = "'" +| from |+ "' to '" +| entryTo ^. path |+ "':"
+          let errorMsg = createErrorToBuilder err
+          pure $ unlinesF @_ @Builder $ header : [indentF 2 errorMsg]
--- a/lib/Backend/Commands.hs
+++ b/lib/Backend/Commands.hs
@@ -23,6 +23,10 @@ import Data.List.NonEmpty (NonEmpty)
 import qualified Data.List.NonEmpty as NE
 import Data.Set (Set)
 import qualified Data.Set as Set
+import qualified Data.List as List
+import Polysemy.Error (throw, Error, runError)
+import Control.Monad.Extra (whenM)
+import qualified Data.List.NonEmpty as NE
 
 import Backend (BackendEffect, listSecrets, readSecret, writeSecret, deleteSecret)
 import Entry (Entry, Field, FieldKey, value, fields, dateModified, newEntry, newField, visibility, FieldVisibility(..), EntryTag, path)
@@ -34,6 +38,7 @@ import qualified Coffer.Path as Path
 import Error (CofferError(..))
 import Coffer.Util (catchAndReturn)
 import Data.Either (rights)
+import Data.List (inits)
 
 runCommand
   :: (Member BackendEffect r, Member (Embed IO) r, Member (Error CofferError) r)
@@ -318,6 +323,8 @@ buildCopyOperations oldPath newPath force = do
       Left entry -> copyEntry entry
       Right dir -> copyDir dir
 
+  checkForEntriesInPath newPath
+
   -- Checks that the paths we're trying to copy entries to aren't already occupied
   -- by an existing directory.
   checkCopyOperations operations
@@ -389,6 +396,14 @@ copyCmd (CopyOptions dryRun oldPath newPath force) = do
   unless dryRun do
     runCopyOperations operations
   pure $ CPRSuccess $ getOperationPaths <$> operations
+
+    checkForEntriesInPath :: Path -> Sem r ()
+    checkForEntriesInPath path =
+      when (not (null (path ^. pathSegments))) do
+        let pathInits = Path.EntryPath . NE.fromList <$> (tail . inits) (path ^. pathSegments)
+        filterM pathIsEntry pathInits >>= \case
+          [] -> pure ()
+          (entry : _)  -> throw $ CPRDestinationHasEntry entry
 
 deleteCmd
   :: (Member BackendEffect r, Member (Error CofferError) r, Member (Error DeleteResult) r)
--- a/lib/Backend/Commands.hs
+++ b/lib/Backend/Commands.hs
@@ -18,15 +18,10 @@ import Data.Time.Calendar.Compat (pattern YearMonthDay)
 import Data.Time.Calendar.Month.Compat (pattern MonthDay)
 import Control.Lens (view)
 import Polysemy.Error (throw, Error)
-import Control.Monad.Extra (whenM, whenJust)
-import Data.List.NonEmpty (NonEmpty)
+import Control.Monad.Extra (whenM)
 import qualified Data.List.NonEmpty as NE
 import Data.Set (Set)
 import qualified Data.Set as Set
-import qualified Data.List as List
-import Polysemy.Error (throw, Error, runError)
-import Control.Monad.Extra (whenM)
-import qualified Data.List.NonEmpty as NE
 
 import Backend (BackendEffect, listSecrets, readSecret, writeSecret, deleteSecret)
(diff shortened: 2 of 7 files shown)


app/Main.hs
```diff
@@ -109,6 +109,8 @@ main = do
             let header = "The following entries cannot be copied because a directory already exists at the destination."
             let errorMsgs = paths <&> \(from, to) -> "Cannot copy '" +| from |+ "' to '" +| to |+ "'."
             printError $ unlinesF @_ @Builder $ header : "" : errorMsgs
+          CPRDestinationHasEntry entry -> printError $ 
+            "The following entries cannot be copied because a destination directory has an entry '" +| build entry |+ "' in it's subpath."
```
dcastro, line 113: > The following entries

Which?

lib/CLI/Types.hs
```diff
@@ -56,6 +56,7 @@ data CopyResult
   | CPRPathNotFound Path
   | CPRMissingEntryName
   | CPRDestinationIsDirectory [(EntryPath, EntryPath)]
+  | CPRDestinationHasEntry EntryPath
```
dcastro, line 59: `CPRDestinationIsDirectory` and `CPREntryAlreadyExists` store a list with all the problematic entries/paths.

I think `CPRDestinationHasEntry` should do the same.

lib/Backend/Commands.hs
```diff
@@ -266,6 +268,7 @@ copyCmd (CopyOptions oldPath newPath force) = do
       Right dir -> copyDir dir
 
   -- Validation checks.
+  checkForEntriesInPath newPath
```
dcastro, line 271: Given this tree:
```
$ cabal run exe:coffer -- create test/dir/entry1                                        
$ cabal run exe:coffer -- create test/dir/entry2
$ cabal run exe:coffer -- view /

/
  test/
    dir/
      entry1 - [2022-02-24 11:25:51]
      entry2 - [2022-02-24 11:25:53]
```

Previously, when you tried to copy `entry1` on top of `entry2`, you'd get this message:

```
$ cabal run exe:coffer -- copy test/dir/entry1 test/dir/entry2

[ERROR] The following entries cannot be copied because an entry already exists at the destination.
Use '--force' or '-f' to overwrite existing entries.

Cannot copy '/test/dir/entry1' to '/test/dir/entry2'.
```

But now, after these changes, the same command returns:
```
[ERROR] The following entries cannot be copied because a destination directory has an entry '/test/dir/entry2' in it's subpath.
```

And even with `-f`, the entry cannot be copied.

lib/CLI/Types.hs
dcastro, line 59: How about renaming `CPRDestinationHasEntry` to `CPRParentDirectoryIsEntry`?

To me, the name "destination has entry" implies that we found an entry at the destination path. E.g., we tried to copy `a/b/c` to `e/f/g/`, but an entry already exists at `e/f/g`. But that's not what this error means.

"Parent directory is entry" is a better fit, in my opinion.

lib/Backend/Commands.hs
```diff
@@ -324,6 +327,14 @@ copyCmd (CopyOptions oldPath newPath force) = do
           <&> (\(CopyOperation old new) -> (old ^. E.path, new ^. E.path) )
           & throw . CPREntryAlreadyExists
 
+    checkForEntriesInPath :: Path -> Sem r ()
+    checkForEntriesInPath path =
+      when (not (null (path ^. unPath))) do
+        let pathInits = Path.EntryPath . NE.fromList <$> (tail . inits) (path ^. unPath)
```
dcastro, line 333: This check is not enough.

Here's a counterexample:

```
$ cabal run exe:coffer -- create test/dir/entry1                                        
$ cabal run exe:coffer -- create test/dir/entry2
$ cabal run exe:coffer -- create test2/dir

$ cabal run exe:coffer -- view /

/
  test/
    dir/
      entry1 - [2022-02-24 11:48:52]
      entry2 - [2022-02-24 11:48:56]
  test2/
    dir - [2022-02-24 11:49:07]
```

```
$ cabal run exe:coffer -- copy test test2
[SUCCESS] Copied '/test/dir/entry1' to '/test2/dir/entry1'.
[SUCCESS] Copied '/test/dir/entry2' to '/test2/dir/entry2'.

$ cabal run exe:coffer -- view /
/
  test/
    dir/
      entry1 - [2022-02-24 11:48:52]
      entry2 - [2022-02-24 11:48:56]
  test2/
    dir - [2022-02-24 11:49:07]
    dir/
      entry1 - [2022-02-24 11:49:20]
      entry2 - [2022-02-24 11:49:20]
```

Now `/test2/dir` points to both a directory and an entry.

dcastro (changes requested):

lib/Backend/Commands.hs
DK318: Done

app/Main.hs

lib/CLI/Types.hs
DK318: Done

lib/Backend/Commands.hs

lib/CLI/Types.hs
DK318: Done

lib/Backend/Commands.hs
```diff
@@ -329,6 +339,70 @@ copyCmd (CopyOptions oldPath newPath force) = do
...
+              checkForEntriesInSubdirs dir1 dir2
+            else
+              pure False
+          else
+            pure False
+
+        checkForEntriesInSubdirs :: Directory -> Directory -> Sem r Bool
+        checkForEntriesInSubdirs dir1 dir2 = do
+          -- 'subdirs' hashmap at first level contains start directory
+          entries <- execStateT (getEqualDirectoriesAndEntries (head . HashMap.elems $ dir1 ^. Dir.subdirs) (head . HashMap.elems $ dir2 ^. Dir.subdirs)) []
```
dcastro, line 386: `head` is an unsafe function, can we avoid using it?

dcastro (changes requested):

lib/Backend/Commands.hs
```diff
@@ -329,6 +339,70 @@ copyCmd (CopyOptions oldPath newPath force) = do
...
+          & throw . CPRParentDirectoryIsEntry
+
+    checkForEntriesInSubpaths :: [CopyOperation] -> Sem r ()
+    checkForEntriesInSubpaths operations = do
+      invalidOperations <- flip filterM operations \(CopyOperation old new) -> do
+        checkForEntriesInSubpaths_ (Path.entryPathAsPath $ old ^. E.path) (Path.entryPathAsPath $ new ^. E.path)
+      when (not (null invalidOperations)) do
+        invalidOperations
+          <&> (\(CopyOperation old new) -> (old ^. E.path, new ^. E.path) )
+          & throw . CPRSubdirectoriesContainEntries
```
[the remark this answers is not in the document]
DK318: Test `entry-directory copy conflict` covers it.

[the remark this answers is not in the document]
dcastro: That test expects the following error message:

> The following entries cannot be copied because a destination directory has an entry in it's subpath.

But `CPRSubdirectoriesContainEntries` prints the following error message:

> The following entries cannot be copied because a directory has an entry in it's subpath.

lib/Backend/Commands.hs
```diff
@@ -329,6 +339,70 @@ copyCmd (CopyOptions oldPath newPath force) = do
     getOperationPaths :: CopyOperation -> (EntryPath, EntryPath)
     getOperationPaths (CopyOperation old new) = (old ^. E.path, new ^. E.path)
 
+    checkForEntriesInPath :: [CopyOperation] -> Sem r ()
+    checkForEntriesInPath operations = do
+      invalidOperations <- flip filterM operations \(CopyOperation _ new) -> do
```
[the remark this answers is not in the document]

lib/Backend/Commands.hs
```diff
@@ -329,6 +339,70 @@ copyCmd (CopyOptions oldPath newPath force) = do
...
+          <&> (\(CopyOperation old new) -> (old ^. E.path, new ^. E.path) )
+          & throw . CPRSubdirectoriesContainEntries
+      where
+        checkForEntriesInSubpaths_ :: Path -> Path -> Sem r Bool
+        checkForEntriesInSubpaths_ path1 path2 = do
+          mDir1 <- getEntryOrDir path1
+          mDir2 <- getEntryOrDir path2
+          if _Just `is` mDir1 && _Just `is` mDir2 then do
+            let eDir1 = mDir1 ^?! _Just
+                eDir2 = mDir2 ^?! _Just
```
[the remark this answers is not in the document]

lib/Backend/Commands.hs
```diff
@@ -24,16 +24,23 @@ import Data.List.NonEmpty (NonEmpty)
 import qualified Data.List.NonEmpty as NE
```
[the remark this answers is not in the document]
DK318: Done

lib/Backend/Commands.hs
```diff
@@ -329,6 +339,70 @@ copyCmd (CopyOptions oldPath newPath force) = do
...
 
+    checkForEntriesInPath :: [CopyOperation] -> Sem r ()
+    checkForEntriesInPath operations = do
+      invalidOperations <- flip filterM operations \(CopyOperation _ new) -> do
+        let path = Path.entryPathAsPath $ new ^. E.path
+        if length (path ^. pathSegments) > 1 then do
+          let pathInits = Path.EntryPath . NE.fromList <$> (init . tail . inits) (path ^. pathSegments)
+          filterM pathIsEntry pathInits >>= \case
+            [] -> pure False
+            _  -> pure True
```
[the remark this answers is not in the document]

lib/Backend/Commands.hs

lib/Backend/Commands.hs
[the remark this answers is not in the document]
DK318: Hmm. I think this check is redunant, because now we're checking `CopyOperations`, which already have all necessary entries.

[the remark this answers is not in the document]

lib/Backend/Commands.hs
```diff
@@ -269,6 +269,19 @@ copyCmd (CopyOptions oldPath newPath force) = do
...
+  -- Checks that the paths we're trying to copy entries to aren't already occupied
+  -- by existing entries in theirs parent directories.
+  checkCopyOperations operations
+    do \(CopyOperation _ new) -> do
+        let entryPath = new ^. E.path
+        let mPathInits = traverse NE.nonEmpty $ (init . NE.tail . NE.inits) (entryPath ^. pathSegments)
+        case mPathInits of
+          Just pathInits -> do
+            let entryPaths = view (from (pathSegments @EntryPath)) <$> pathInits
+            anyM pathIsEntry entryPaths
```
dcastro, line 281: In #34, where we do a check similar to this, we report the parent directory(ies) where an entry was found. We should do the same here.

In fact, now that I think about it, the checks performed in `createCmd` and `copyCmd` are the same. Conceptually, it makes sense: both operations involve creating a new entry.

So it follows that maybe we should have a single function that performs all 3 checks to make sure an entry can be created. And that function could be reused in both `createCmd` and `copyCmd`.

(BTW, this PR should probably be based on top of #34 because you'll need to reuse some stuff from there)

dcastro (changes requested):

app/Main.hs
```diff
@@ -114,6 +114,10 @@ main = do
             let header = "The following entries cannot be copied because a directory already exists at the destination."
```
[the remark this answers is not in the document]
DK318: I've renamed those commits. Is it enough?

lib/Backend/Commands.hs

app/Main.hs
```diff
@@ -116,6 +116,10 @@ main = do
...
             let errorMsgs = NE.toList paths <&> \(from, to) -> "Cannot copy '" +| from |+ "' to '" +| to |+ "'."
             printError $ unlinesF @_ @Builder $ header : "" : errorMsgs
+          CPRParentDirectoryIsEntry paths -> do
+            let header = "The following entries cannot be copied because a destination directory has an entry in it's subpath."
+            let errorMsgs = NE.toList paths <&> \(from, to) -> "Cannot copy '" +| from |+ "' to '" +| to |+ "'."
```
Heimdell, line 121: I recommend https://hackage.haskell.org/package/nyan-interpolation-0.9/docs/Text-Interpolation-Nyan.html

Heimdell (approved):

app/Main.hs
dcastro: Oh good point! I'll open an issue for this.

dcastro: Done: https://github.com/serokell/coffer/issues/39

tests/golden/copy-command/copy-command.bats
```diff
@@ -174,3 +174,56 @@ EOF
...
+EOF
+
+  run coffer copy /a /x -f
+
+  assert_failure
+  assert_output - <<EOF
+[ERROR] The following entries cannot be copied because a directory already exists at the destination.
+
+Cannot copy '/a/b' to '/x/b'.
+EOF
```
dcastro, line 198: Since we're executing two commands in quick succession and we expect them both to return the exact same output, I think it makes sense to use a variable to store the expected output:

```suggestion
  EXPECTED=$(cat <<EOF
[ERROR] The following entries cannot be copied because a directory already exists at the destination.

Cannot copy '/a/b' to '/x/b'.
EOF
)
  run coffer copy /a /x
  assert_failure
  assert_output "$EXPECTED"

  run coffer copy /a /x -f
  assert_failure
  assert_output "$EXPECTED"
```

app/Main.hs
```diff
@@ -116,6 +116,10 @@ main = do
             let header = "The following entries cannot be copied because a directory already exists at the destination."
             let errorMsgs = NE.toList paths <&> \(from, to) -> "Cannot copy '" +| from |+ "' to '" +| to |+ "'."
             printError $ unlinesF @_ @Builder $ header : "" : errorMsgs
+          CPRParentDirectoryIsEntry paths -> do
+            let header = "The following entries cannot be copied because a destination directory has an entry in it's subpath."
```
dcastro, line 120:
```suggestion
            let header = "The following entries cannot be copied because a destination directory has an entry in its subpath."
```

dcastro (changes requested):

lib/Backend/Commands.hs
dcastro: > So it follows that maybe we should have a single function that performs all 3 checks to make sure an entry can be created. And that function could be reused in both createCmd and copyCmd.

I don't think this part is done yet.

DK318: I don't know how to do it. We want this function to return smth like `Sem r CheckResult` in order not to lose the result of the check. But in `copyCmd` it would be difficult to deal with this list.

tests/golden/copy-command/copy-command.bats
DK318: Done

app/Main.hs

lib/Backend/Commands.hs

lib/Backend/Commands.hs
```diff
@@ -482,3 +478,45 @@ getEntryOrDir path =
...
+        -- Ignore paths that cannot possibly point to an entry
+        <&> Path.pathAsEntryPath
+        & rights
+
+  filterM pathIsEntry parentDirsExceptRoot
+
+checkCreateEntry
+  :: forall r
+   . (Member BackendEffect r)
+  => Bool -> Entry -> Sem r CreateCheckResult
```
dcastro, line 500: I think returning a `Validation [CreateError] Entry` would be semantically better.

If you're not familiar with it: [`Validation e a`](https://hackage.haskell.org/package/either-5.0.1.1/docs/Data-Either-Validation.html#t:Validation) is very similar to `Either e a`, except:
1. `Validation` has an applicative instance but no monad instance.
2. When you `sequence` many `Either`s, you only get the first error and all the other items are ignored. On the other hand, when you `sequenceA` many `Validation`s, all the errors all collated together (which is the behaviour we want here).

After changing `checkCreateEntry` to return `Sem r (Validation [CreateError] Entry)`, you can collate all the errors using `for + sequenceA`:

```hs
do
  let entries = [] :: [Entry]
  let force = True
 
  sequenceA <$> for entries (checkCreateEntry2 force) >>= \case
    Failure allErrors -> pure ()
    Success allEntries -> pure ()
```

dcastro (changes requested):

lib/Backend/Commands.hs
```diff
@@ -295,17 +283,25 @@ copyCmd (CopyOptions oldPath newPath force) = do
       Left entry -> copyEntry entry
       Right dir -> copyDir dir
 
+  processedOperations <- forM (operations <&> coNew) (checkCreateEntry force) <&> zip operations
```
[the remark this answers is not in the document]

lib/Backend/Commands.hs

app/Main.hs
```diff
@@ -120,17 +124,24 @@ main = do
...
+                                      ]
+                                pure $ unlinesF @_ @Builder $ header : [indentF 2 errorMsg]
+                              (from, CEDestinationIsDirectory entryTo) -> do
+                                let header = "'" +| from |+ "' to '" +| entryTo ^. path |+ "':"
+                                let errorMsg = "'" +| entryTo ^. path |+ "' is a directory."
+                                pure $ unlinesF @_ @Builder $ header : [indentF 2 errorMsg]
+                              (from, CEParentDirectoryIsEntry (entryTo, clashed)) -> do
+                                let header = "'" +| from |+ "' to '" +| entryTo ^. path |+ "':"
+                                let errorMsg = "Attempted to create the directory '" +| clashed |+ "' but an entry exists at that path."
+                                pure $ unlinesF @_ @Builder $ header : [indentF 2 errorMsg]
```
dcastro, line 143: All 3 branches follow the exact same template for building the message, the only thing that changes is the `errorMsg`. So maybe this could be refactored to avoid this duplication.

lib/Backend/Commands.hs
```diff
@@ -316,20 +289,13 @@ buildCopyOperations oldPath newPath force = do
...
-      do \(CopyOperation _ new) -> pathIsEntry (new ^. E.path)
-      do CPREntryAlreadyExists . fmap getOperationPaths
+  sequenceA <$> forM operations validateCopyOperation >>= \case
+    Failure createErrors -> throw $ CPRCreateErrors createErrors
+    Success _ -> pure operations
 
-  pure operations
+  -- sequenceA <$> forM (operations <&> coNew) (checkCreateEntry force) >>= \case
+  --   Failure createErrors -> throw $ CPRCreateErrors createErrors
+  --   Success _ -> pure operations
```
dcastro, line 298: Forgot to delete this commented out code.

lib/Backend/Commands.hs
```diff
@@ -360,15 +326,8 @@ buildCopyOperations oldPath newPath force = do
       CopyOperation old (new & dateModified .~ nowUtc)
 
     -- | Performs a check on all `CopyOperation`s and throws an error if any of checks fail.
```
dcastro, line 328: Since you changed what this function does, this haddock is now outdated.

lib/Backend/Commands.hs
```diff
@@ -360,15 +326,8 @@ buildCopyOperations oldPath newPath force = do
...
-      :: [CopyOperation]
-      -> (CopyOperation -> Sem r Bool)
-      -> (NonEmpty CopyOperation -> CopyResult)
-      -> Sem r ()
-    checkCopyOperations operations pred mkErr = do
-      invalidOperations <- filterM pred operations
-      whenJust (NE.nonEmpty invalidOperations) \invalidOperationsNE ->
-        throw $ mkErr invalidOperationsNE
+    validateCopyOperation :: CopyOperation -> Sem r (Validation [(EntryPath, CreateError)] Entry)
+    validateCopyOperation (CopyOperation old new) = checkCreateEntry force new <&> first \err -> [(old ^. path, err)]
```
dcastro, line 330:
```suggestion
    validateCopyOperation (CopyOperation old new) =
      checkCreateEntry force new <&> first \err -> [(old ^. path, err)]
```

dcastro (changes requested):
Great job :+1: just needs some final cleaning up, and we're good to go!

app/Main.hs
```diff
@@ -120,17 +124,24 @@ main = do
...
-              CPREntryAlreadyExists paths -> do
-                let header = unlinesF @_ @Builder
-                      [ "The following entries cannot be renamed because an entry already exists at the destination."
-                      , "Use '--force' or '-f' to overwrite existing entries."
-                      ]
-                let errorMsgs = NE.toList paths <&> \(from, to) -> "Cannot rename '" +| from |+ "' to '" +| to |+ "'."
-                printError $ unlinesF @_ @Builder $ header : "" : errorMsgs
+              CPRCreateErrors errors -> do
+                errorMsgs <- forM errors \case
+                              (from, CEEntryAlreadyExists entryTo) -> do
```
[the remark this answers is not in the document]
DK318: Don't get how can I reuse it in `create`. We don't have `from` there.

lib/Backend/Commands.hs

lib/Backend/Commands.hs

lib/Backend/Commands.hs

app/Main.hs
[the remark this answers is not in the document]
dcastro: Oh yeah good point!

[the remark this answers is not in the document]
dcastro: I think we can still have a function `CreateError -> Builder` that maps the error to just the error message (e.g. `CEDestinationIsDirectory entryTo` to `"/path" is a directory".

app/Main.hs
dcastro: Doesn't look like it's done :/ Also see my comment above about having a function dedicated to building the `errorMsg` part, and reusing that function for `create` + `copy/rename`.

dcastro (approved):
One last minor thing, i'm approving it now anyway! :+1:

Issue #15, opened 2022-02-23
Move config to a config file


dcastro:
@MagicRB has written a `Config` module with the relevant data types & decoders.

In the `diogo/cli` branch, we are still hardcoding the configuration in `Main`.

We should refactor the CLI so that the config is read from a file rather than being hardcoded.

It should also be noted that the `Config` data types support multiple backends, but the CLI assumes there's only 1 backend. The CLI will have to be changed to support multiple backends, but this can be done in a separate issue (#16). For now, let's just assume the user intends to use the first backend from the config file.

PR #17, opened 2022-02-23
[#15] Backends config implementation

DK318:
Problem: there was no config file for backends. All backend stuff should be hardcoded in order to use `coffer`.
Solution: used @MagicRB data types and functions from `Config` module.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -8,36 +8,31 @@ import Control.Lens
 import Control.Monad (forM_)
 import Options.Applicative ( execParser )
 import Polysemy
-import Servant.Client (BaseUrl(..), Scheme(Http))
 import Polysemy.Error (errorToIOFinal, Error)
-import Data.Text (Text)
-import System.Exit (die)
+import System.Exit (die, exitFailure)
 import Fmt
+import qualified Data.Text.IO as TIO
+import qualified Toml
+import qualified Data.HashMap.Strict as HS
 import qualified Data.List.NonEmpty as NE
 
 import CLI.Parser
 import CLI.Types
 import CLI.PrettyPrint
-import Backend.Vault.Kv
 import Error
 import Backend.Commands as Commands
 import Backend
 import Coffer.Path ( Path, EntryPath )
 import qualified Entry as E
 import qualified Coffer.Directory as Dir
+import Config (configCodec, Config (..))
 
-url :: BaseUrl
-url = BaseUrl Http "127.0.0.1" 8200 ""
-
-token :: VaultToken
-token = VaultToken "s.AnHJuzOxRiD12eLxALFytccC"
-
-mount :: Text
-mount = "secret"
-
-runVaultIO_ :: Sem '[BackendEffect, Error CofferError, Embed IO, Final IO ] a -> IO a
-runVaultIO_ action =
-  runVaultIO url token mount action
+runBackend
+  :: SomeBackend
+  -> Sem '[BackendEffect, Error CofferError, Embed IO, Final IO ] a
+  -> IO a
+runBackend (SomeBackend backend) action =
+  _runEffect backend action
     & errorToIOFinal @CofferError
     & embedToFinal @IO
     & runFinal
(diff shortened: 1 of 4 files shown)


MagicRB:
Hope the API I created isn't too bad :D

config.toml
```diff
@@ -0,0 +1,8 @@
+main_backend = "main"
+
+[[backend]]
+  type = "vault-kv"
+  name = "main"
+  address = "127.0.0.1:8200"
+  mount = "secret"
+  token = "s.XxzTITDlzOsNNzBlYRHURna6"
```
dcastro, line 8: The default config file shouldn't have any specific token. I think a placeholder like `"<vault token>"` would do.

config.toml
```diff
@@ -0,0 +1,8 @@
+main_backend = "main"
+
+[[backend]]
```
dcastro, line 3: Why the double square brackets `[[` and `]]`?

config.toml
```diff
@@ -0,0 +1,8 @@
+main_backend = "main"
```
dcastro, line 1: This file should be added to the `extra-source-files` section in `package.yaml`.

dcastro (changes requested):

config.toml
MagicRB: Because it's a list backends, in haskell it creates `[Backend]`

dcastro: Gotcha, thanks!

app/Main.hs
```diff
@@ -11,8 +11,11 @@ import Polysemy
 import Servant.Client (BaseUrl(..), Scheme(Http))
```
dcastro, line 11: Some of these imports are redundant.
Please run the pipeline against your PR.

config.toml
dcastro: Why was `config.toml` deleted?

app/Main.hs
dcastro, line 11: The file `test.toml` should be deleted (or rather, replaced with/renamed to `config.toml`)

app/Main.hs
```diff
@@ -46,104 +44,118 @@ runVaultIO_ action =
       Left err -> do
         die $ show err
 
+readConfig :: IO Config
+readConfig = do
+  text <- TIO.readFile "./config.toml"
+  let config = Toml.decode configCodec text
+  case config of
+    Left err -> TIO.putStrLn (Toml.prettyTomlDecodeErrors err) >> exitFailure
```
dcastro, line 52: Please also print a header line like `An error occurred while decoding <config filename>:`

app/Main.hs
```diff
@@ -25,19 +28,14 @@ import Backend
...
-
-mount :: Text
-mount = "secret"
-
-runVaultIO_ :: Sem '[BackendEffect, Error CofferError, Embed IO, Final IO ] a -> IO a
-runVaultIO_ action =
-  runVaultIO url token mount action
+import Config (configCodec, Config (..))
+
+runVaultIO_
```
dcastro, line 33: I guess this function has now been generalized to work with _any_ backend (not necessarily vault's backend), so maybe we should rename it to something more generic like `runBackend`.

dcastro (changes requested):

config.toml
DK318: Done

config.toml
DK318: Done

app/Main.hs

app/Main.hs

app/Main.hs

app/Main.hs
```diff
@@ -46,104 +41,118 @@ runVaultIO_ action =
       Left err -> do
         die $ show err
 
+readConfig :: IO Config
+readConfig = do
+  text <- TIO.readFile "./config.toml"
+  let config = Toml.decode configCodec text
+  case config of
+    Left err -> TIO.putStrLn ("An error occurred while decoding 'config.toml':" <> Toml.prettyTomlDecodeErrors err) >> exitFailure
```
dcastro, line 49: Please print the header msg ("An error occurred ...") and the actual error in separate lines.

If they're printed in the same line, and the actual error msg spans multiple lines, it's gonna look weird.

```suggestion
    Left err -> do
      TIO.putStrLn "An error occurred while decoding 'config.toml':"
      TIO.putStrLn $ Toml.prettyTomlDecodeErrors err
      exitFailure
```

dcastro (approved):

Issue #16, opened 2022-02-23
Support multiple backends from the CLI


dcastro:
This issue depends on #15.

Right now, the CLI assumes there's only 1 backend. It will need to be refactored to allow the user to specify which backend they want to interact with.

@MagicRB has suggested the following syntax:

```
coffer view mybackend#some/path
```

DK318:
Maybe it would be better to have syntax like this?
```
coffer #mybackend view some/path
```
Because it is not clear how to support this for e.g. `find` command.

Heimdell:
`-b mybackend` or `--backend mybackend`?

MagicRB:
Why not `coffer find backend# search_term`?

MagicRB:
> Or `coffer view /backend/some/path`, a la "absolute path"

I think that avoiding paths that look like filesystem path but aren't filesystem path us for the best.

Heimdell:
I suggest that we either use shell and don't do risky stuff or make it a repl-like and do whatever we want.

MagicRB:
Well nix uses essentially the same syntax i proposed and its fine imo

dcastro:
Looks like we have two options, either specify the backend as a separate argument (`-b mybackend`) or as part of the path argument (`backend#path`, `backend/path`, etc.)

I think there's a question we need to consider first, and I believe the answer will lead us down the right path:

Do we want to allow users to copy/move entries from one backend to another?

Remember that the `copy`/`rename` commands take two paths as arguments.

So if we want to allow copying/renaming from one backend to another, then I think specifying the backend in the path would be more ergonomic:

```
coffer copy vault#/some/path pass#/other/path
```

If we _don't_ want to allow users to do this, then `-b` makes more sense:

```
coffer copy -b vault /some/path /other/path
```

I think we should allow users to do this (I don't see a reason not to). What do you all think? /cc @notgne2

MagicRB:
Personally, as a future user, I really need to be able to copy between backends. I use pass for my own personal secrets but Vault for machine secrets (I utilize HashiCorp Nomad and Consul) so having the ability to quickly copy a secret from one to the other is really desirable for me. Especially since we manage to retain at least read-only compatibility with normal Vault tooling.

dcastro:
Regarding `backend#path` vs `backend/path`, again I think it boils down to this:

[...]

PR #18, opened 2022-02-23
[#16] Multiple backends support

DK318:
Problem: at this point we can store secrets only in one backend.
Solution: added optional argument [-b|--backend] which specifies the backend that we are working with.
Fixed: #16

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -55,13 +55,16 @@ readConfig = do
 main :: IO ()
 main = do
   config <- readConfig
-  let mainBackendName = mainBackend config
-      mBackend = HS.lookup mainBackendName (backends config)
+  let defaultBackend = mainBackend config
+  options <- execParser (parserInfo defaultBackend)
+  let someCommand = _someCommand options
+      backendName = _backend options
+      mBackend = HS.lookup backendName (backends config)
   case mBackend of
-    Nothing -> TIO.putStrLn ("Can't find backend '" <> mainBackendName <> "' in 'config.toml'") >> exitFailure
+    Nothing -> TIO.putStrLn ("Can't find backend '" <> backendName <> "' in 'config.toml'") >> exitFailure
     Just packedBackend -> do
       runBackend packedBackend do
-        embed (execParser parserInfo) >>= \case
+        case someCommand of
           SomeCommand cmd@CmdView{} -> do
             runCommand cmd >>= \case
               VRDirectory dir -> pprint $ buildDirectory dir
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -13,11 +13,11 @@ import System.Exit (die, exitFailure)
 import Fmt
 import qualified Data.Text.IO as TIO
 import qualified Toml
-import qualified Data.HashMap.Strict as HS
 
 import CLI.Parser
 import CLI.Types
 import CLI.PrettyPrint
+import Backend.Interpreter
 import Error
 import Backend.Commands as Commands
 import Backend
@@ -27,12 +27,11 @@ import qualified Coffer.Directory as Dir
 import Config (configCodec, Config (..))
 import Entry (path, Entry)
 
-runBackend
-  :: SomeBackend
-  -> Sem '[BackendEffect, Error CofferError, Embed IO, Final IO ] a
+runBackendIO
+  :: Sem '[BackendEffect, Error CofferError, Embed IO, Final IO ] a
   -> IO a
-runBackend (SomeBackend backend) action =
-  _runEffect backend action
+runBackendIO action =
+  runBackend action
     & errorToIOFinal @CofferError
     & embedToFinal @IO
     & runFinal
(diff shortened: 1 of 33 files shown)


lib/CLI/Types.hs
```diff
@@ -13,6 +13,12 @@ import Coffer.Path (Path, EntryPath)
 import Data.Set (Set)
 import Data.List.NonEmpty ( NonEmpty )
 
+data Options = Options
+  { _backend :: Text 
+  , _someCommand :: SomeCommand
```
dcastro, line 18: Please use the same field naming convention as used in other data types (`oBackend`, `oSomeCommand`)

lib/CLI/Parser.hs
```diff
@@ -38,15 +38,28 @@ import Coffer.Path (Path, mkPath, EntryPath, mkEntryPath)
...
+  where
+    optionalBackendParser :: Parser Text
+    optionalBackendParser = 
+      option str $ mconcat
+        [ long "backend"
+        , short 'b'
+        , metavar "BACKEND"
+        , value defaultBackend
+        , help "Backend with which actions will be performed. If not specified, then main backend would be chosen" 
+        ]
```
dcastro, line 59: I thought the consensus was that the backend should be part of the path (and not a separate cli option)?

dcastro (changes requested):

Heimdell (approved):

Heimdell:
(approve if @dcastro's issues with PR are fixed)

lib/CLI/Parser.hs
DK318: Supported both variants (`[-b|--backend]` for `find` and `rename` and hash-separated for other commands).

dcastro: Why should we support both variants?

DK318: I think it's weird to use `QualifiedPath` in `rename` command because we can perform 
````
coffer rename first#entry1 second#entry2
````

In `find` command path is optional, so we can clarify the backend we are working with without providing the path.

Comment (dcastro) [reply lib/CLI/Parser.hs]: > I think it's weird to use QualifiedPath in rename command because we can perform
> `coffer rename first#entry1 second#entry2`

Why is that weird? It seems ok to me. It's not that different from `coffer copy first#entry1 second#entry2`, which we want to support.

----


> In `find` command path is optional, so we can clarify the backend we are working on without providing the path.

Richard addressed that point here: https://github.com/serokell/coffer/issues/16#issuecomment-1048858199

> Why not `coffer find backend# search_term`?

File: lib/Backend/Interpreter.hs [hunk 3]
```diff
@@ -0,0 +1,18 @@
...
+--
+-- SPDX-License-Identifier: MPL-2.0
+
+module Backend.Interpreter where
+
+import Polysemy
+
+import Backend (Backend (..), BackendEffect (..), SomeBackend (..), Effects)
+
+runBackendIO :: Effects r => Sem (BackendEffect ': r) a -> Sem r a
```
dcastro, line 11: This function doesn't deal with `IO`, so I think it should be renamed to just `runBackend`.

Similarly, the `runBackendIO_` in `Main` could  be renamed to simply `runBackendIO`.

lib/Backend/Commands.hs
```diff
@@ -456,3 +468,11 @@ getEntryOrDir path =
...
                     , "Got: '" <> secret <> "'."
                     ]
+
+getBackend :: forall r. Member (Error CofferError) r => Config -> Maybe BackendName -> Sem r SomeBackend
+getBackend config backendNameMb = do
+  let backendName = maybe (mainBackend config) getBackendName backendNameMb
+  let backendsHashMap = backends config
+  case backendsHashMap !? backendName of
+    Just backend -> pure backend
+    Nothing -> throw BackendNotFound
```
dcastro, line 478: We should mention _which_ backend name was not found.

dcastro (changes requested):
Cool approach, I like it! :smile: :+1:

dcastro:
Don't forget to:
1. Add some tests after rebasing
2. Add `Fixed: <issue number>` to the PR's description

lib/Backend/Interpreter.hs
```diff
@@ -0,0 +1,18 @@
...
+runBackendIO = interpret $
+  \operation -> do
+    case operation of
+      WriteSecret (SomeBackend backend) entry -> _writeSecret backend entry
+      ReadSecret (SomeBackend backend) path -> _readSecret backend path
+      ListSecrets (SomeBackend backend) path -> _listSecrets backend path
+      DeleteSecret (SomeBackend backend) path -> _deleteSecret backend path
```
dcastro, line 18:
```suggestion
runBackendIO = interpret \case
  WriteSecret (SomeBackend backend) entry -> _writeSecret backend entry
  ReadSecret (SomeBackend backend) path -> _readSecret backend path
  ListSecrets (SomeBackend backend) path -> _listSecrets backend path
  DeleteSecret (SomeBackend backend) path -> _deleteSecret backend path
```

dcastro (changes requested):

app/Main.hs
```diff
@@ -55,107 +54,104 @@ readConfig = do
 main :: IO ()
 main = do
```
dcastro, line 55: We should make sure paths appear qualified in both success and error messages.

```
coffer copy 'b1#/a' 'b2#/a'  
[SUCCESS] Copied '/a/b' to '/a/b'.
```

dcastro (changes requested):

app/Main.hs
DK318: I think we should create a separate issue for this because after fixing this many `bats` tests would fail. And it would be more comfortable to write tests if we merge #18 and #21.

lib/CLI/Parser.hs
DK318: Done

lib/CLI/Parser.hs
```diff
@@ -277,21 +293,25 @@ tagOptions =
...
-      ]
+readEntryPath' :: Text -> Either String EntryPath
+readEntryPath' input =
+  mkEntryPath input & first \err -> unlines
+    [ "Invalid entry path: '" <> T.unpack input <> "'."
+    , T.unpack err
+    ]
+
+_readEntryPath :: ReadM EntryPath
+_readEntryPath = str >>= toReader . readEntryPath'
```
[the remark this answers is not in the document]
DK318: Done

lib/Backend/Interpreter.hs
DK318: Done

lib/Backend/Commands.hs

lib/Backend/Interpreter.hs
DK318: Done

app/Main.hs
```diff
@@ -13,12 +13,12 @@ import System.Exit (die, exitFailure)
 import Fmt
```
dcastro, line 13: The latest fixup commit will be squashed against `[#16] `BackendName` newtype add`, but it contains changes that relate to 3 different commits. 

This fixup commit should be split accordingly.

lib/Error.hs
```diff
@@ -11,5 +11,6 @@ import qualified Data.Text as T
 data CofferError =
   MarshallingFailed
   | ConnectError
+  | BackendNotFound T.Text
```
dcastro, line 14: These 3 definitions should be changed from `Text` to `BackendName`:

* `CofferError.BackendNotFound`
* `Config.mainBackend`
* `Backend._name`

lib/Backend/Commands.hs
```diff
@@ -176,11 +185,13 @@ deleteFieldCmd (DeleteFieldOptions path fieldName) = do
...
                 & dateModified .~ nowUtc
-          void $ writeSecret newEntry
+          void $ writeSecret backend newEntry
           pure $ DFRSuccess newEntry
 
-findCmd :: (Member BackendEffect r, Member (Error CofferError) r) => FindOptions -> Sem r (Maybe Directory)
-findCmd (FindOptions pathMb textMb sortMb filters filterFields) = do
+findCmd :: (Member BackendEffect r, Member (Error CofferError) r) => Config -> FindOptions -> Sem r (Maybe Directory)
+findCmd config (FindOptions qPathMb textMb sortMb filters filterFields) = do
+  let backendNameMb = maybe ((eitherToMaybe . newBackendName) $ mainBackend config) qpBackendName qPathMb
```
dcastro, line 193: No need to default to use `maybe + mainBackend` here. This is already done in `getBackend`.

dcastro (changes requested):

lib/Backend/Interpreter.hs
dcastro: I think you forgot this bit:

> Similarly, the runBackendIO_ in Main could be renamed to simply runBackendIO.

app/Main.hs
dcastro: Sure, can you please create an issue for that?

DK318: Created #52

dcastro: Looks great, thanks!

lib/Backend/Interpreter.hs
DK318: Done

app/Main.hs
DK318: Done

lib/Error.hs
DK318: Done

lib/Backend/Commands.hs

lib/BackendName.hs
```diff
@@ -0,0 +1,64 @@
...
+_BackendName = Toml.BiMap textToBackendName backendNameToText
+  where
+    textToBackendName :: T.Text -> Either Toml.TomlBiMapError BackendName
+    textToBackendName t = first Toml.ArbitraryError (newBackendName t)
+
+    backendNameToText :: BackendName -> Either Toml.TomlBiMapError T.Text
+    backendNameToText = Right . getBackendName
+
+backendNameCodec :: Toml.Key -> Toml.TomlCodec BackendName
+backendNameCodec backendNameKey = Toml.Codec input output
```
dcastro, line 52: We can simply this implementation by reusing `Toml.text` codec, and then mapping back/forth over it (using the `didimatch` function defined in `Backend.Vault.Kv`):

```hs
backendNameCodec :: Toml.Key -> Toml.TomlCodec BackendName
backendNameCodec backendNameKey =
  didimatch
    (Just . getBackendName)
    (eitherToMaybe . newBackendName)
    (Toml.text backendNameKey)
```

dcastro: Also:
1. we should move `didimatch` to somewhere where it can be reused, like `Coffer.Util`
2. we should refactor `didimatch` to use `Either Text x` instead of `Maybe x`, so we can report _why_ a conversion failed.

dcastro (changes requested):

lib/BackendName.hs
DK318: Done

lib/Coffer/Util.hs
```diff
@@ -18,3 +24,26 @@ import Polysemy.Error ( runError, Error )
...
+didimatch matchB matchA codec = Toml.Codec
+    { Toml.codecRead = \t -> case Toml.codecRead codec t of
+        Success a ->
+          case matchA a of
+            Left err -> Failure [Toml.ParseError $ Toml.TomlParseError err]
+            Right b -> Success b
+        Failure b -> Failure b
+    , Toml.codecWrite = \c -> case matchB c of
+        Left _ -> empty
+        Right d  -> Toml.codecWrite codec d >>= Toml.eitherToTomlState . matchA
```
dcastro, line 48: We shouldn't be swallowing the error in `Left`. How about this?

```suggestion
    , Toml.codecWrite = \b -> do
        a <- Toml.eitherToTomlState $ matchB b
        a' <- Toml.codecWrite codec a
        Toml.eitherToTomlState $ matchA a'
```

dcastro (changes requested):
Looks like we're only missing some tests now and we're good to go.

lib/Coffer/Util.hs
DK318: Done

dcastro: Looks good!

As a sidenote: turns out the [implementation](https://hackage.haskell.org/package/tomland-1.3.3.1/docs/src/Toml.Codec.Types.html#eitherToTomlState) of `eitherToTomlState` _also_ swallows the error in the `Left` :weary: 

```hs
eitherToTomlState :: Either e a -> TomlState a
eitherToTomlState e = TomlState (either (const Nothing) Just e,)
```

Oh well, seems like the way `TomlState` is represented there's just no way of properly reporting errors in `codecWrite`...

dcastro (approved):

dcastro:
> doesn't moving the effect functions into the `Backend a` kind of beat the point of using polysemy?

Do you mean the ability to have multiple interpreters for the same effect?

We can still have different interpreters:
```hs
logBackend :: Member (Embed IO) r => Sem (BackendEffect ': r) a -> Sem r a
logBackend = interpret \case
  WriteSecret _ entry -> embed do
    putStrLn "writing secret"
    print entry
```

Though we don't have the ability to have an interpreter that discriminates between different types of backends (e.g. an interpreter that behaves one way for `vault` and another way for `pass`).

> I don't think this is a very elegant solution

I agree something _feels_ odd but I can't quite put my finger on it. Do you have an alternative solution?

MagicRB:
We could maybe add an effect, which temporarily switches the interpreter
```haskell
do
  readSecret
  res <- tempSwitch _runBackendPass $ do
    readSecret
    writeSecret
```
Essentially `tempSwitch` is just `embed` actually not essentially but yes it is. What API would we want? Im having a hard time to even imagining a good api

Issue #29, opened 2022-02-28
Implement CLI


dcastro:
Implement a CLI according to [the spec](https://www.notion.so/serokell/Password-sharing-2-232733503bb4469d85713d08f3b3a5e4#2459058a49df42ed84bcc37dd1d6fd4c).

Issue #19, opened 2022-02-24
Allow specifying a different config file


dcastro:
We should support a "global option" (an option that applies to all CLI commands), like `-c/--config`, that allows users to specify a different config file than the default (`./config.toml`).

PR #30, opened 2022-02-28
[#29] Add CLI parser

dcastro:
This PR implements the parsers needed for the CLI and adds a couple of newtypes for type-safety (`PathSegment`, `Path`, `EntryPath`, `FieldVisibility`).

### DIFF
--- a/lib/CLI/Parser.hs
+++ b/lib/CLI/Parser.hs
@@ -1,0 +1,550 @@
+-- SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io>
+--
+-- SPDX-License-Identifier: MPL-2.0
+
+{-# LANGUAGE OverloadedLists #-}
+
+module CLI.Parser
+  ( parserInfo
+  ) where
+
+import Options.Applicative
+import qualified Options.Applicative.Help.Pretty as Pretty
+import Data.Text (Text)
+import qualified Data.Text as T
+import Data.Map (Map)
+import qualified Data.Map as M
+import Data.Functor ((<&>), ($>))
+import qualified Data.List as List
+import Data.Function ((&))
+import Control.Arrow ((>>>))
+import qualified Text.Megaparsec as P
+import qualified Text.Megaparsec.Char as P
+import qualified Text.Megaparsec.Char.Lexer as P
+import Data.Void ( Void )
+import Data.Foldable (asum)
+import Data.Time.Compat (LocalTime (..), makeTimeOfDayValid, localTimeToUTC, utc)
+import Data.Time.Calendar.Compat ( fromGregorianValid )
+import Data.Time.Calendar.Month.Compat (fromYearMonthValid)
+import Control.Monad (guard, void)
+import qualified Data.Char as Char
+import Data.Fixed (Pico)
+import Data.Bifunctor (first)
+import qualified Data.Set as Set
+
+import CLI.Types
+import Entry (FieldKey, EntryTag, newEntryTag, newFieldKey, FieldVisibility (Public, Private))
+import Coffer.Path (Path, mkPath, EntryPath, mkEntryPath)
+
+{-# ANN module ("HLint: ignore Use <$>" :: Text) #-}
+
+parserInfo :: ParserInfo Command
+parserInfo =
+  info (parser <**> helper) $
+    fullDesc
+    <> progDesc "TODO: coffer description goes here"
+    <> header "TODO: coffer description goes here"
+
+parser :: Parser Command
+parser =
+  subparser $ mconcat
+    [ mkCommand "view" CmdView viewOptions
+        "View entries under the specified path, optionally returning only the specified field for each entry"
+    , mkCommand "create" CmdCreate createOptions
+        "Create a new entry at the specified path"
+    , mkCommand "set-field" CmdSetField setFieldOptions
+        "Set a field on the entry at the specified path"
+    , mkCommand "delete-field" CmdDeleteField deleteFieldOptions
+        "Delete a field from the entry at the specified path"
+    , mkCommand "find" CmdFind findOptions
+        "Find and list entries, optionally filtering"
+    , mkCommand "rename" CmdRename renameOptions
+        "Rename an entry or directory"
+    , mkCommand "copy" CmdCopy copyOptions
+        "Copy an entry or directory"
+    , mkCommand "delete" CmdDelete deleteOptions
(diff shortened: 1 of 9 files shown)


lib/CLI/Parser.hs
```diff
@@ -0,0 +1,550 @@
...
+
+readPath :: ReadM Path
+readPath = do
+  input <- str
+  case mkPath input of
+    Right path -> pure path
+    Left err -> readerError $ unlines
+      [ "Invalid path: '" <> T.unpack input <> "'."
+      , T.unpack err
+      ]
```
DK318, line 276: Why not [etherReader](https://hackage.haskell.org/package/optparse-applicative-0.17.0.0/docs/Options-Applicative.html#v:eitherReader)?

lib/CLI/Types.hs
```diff
@@ -0,0 +1,129 @@
...
+  { voPath :: Path
+  , voFieldName :: Maybe FieldKey
+  }
+  deriving stock Show
+
+data CreateOptions = CreateOptions
+  { coPath :: EntryPath
+  , coEdit :: Bool
+  , coForce :: Bool
+  , coTags :: [EntryTag]
```
DK318, line 35: Why not [Data.Set](https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Set.html)?

lib/CLI/Parser.hs
```diff
@@ -0,0 +1,550 @@
...
+import Coffer.Path (Path, mkPath, EntryPath, mkEntryPath)
+
+{-# ANN module ("HLint: ignore Use <$>" :: Text) #-}
+
+parserInfo :: ParserInfo Command
+parserInfo =
+  info (parser <**> helper) $
+    fullDesc
+    <> progDesc "TODO: coffer description goes here"
+    <> header "TODO: coffer description goes here"
```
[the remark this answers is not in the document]
DK318: Also, it can't parse `[-q|--quiet]` and `[-j|--json]`

lib/CLI/Parser.hs
dcastro: Good question, I guess it would be more idiomatic
```hs
  eitherReader \input ->
    case mkPath (T.pack input) of
      Right path -> pure path
      Left err -> Left $ unlines
        [ "Invalid path: '" <> input <> "'."
        , T.unpack err
        ]
```

Could be shortened a bit more using `first` to map over the error:
```hs
  eitherReader \input ->
    mkPath (T.pack input) & first \err ->
      unlines
        [ "Invalid path: '" <> input <> "'."
        , T.unpack err
        ]
```

lib/CLI/Types.hs
dcastro: I'll change this next time I rebase on `magic_rb/core`

lib/CLI/Parser.hs
```diff
@@ -0,0 +1,550 @@
...
+    <*> optional (argument str $ mconcat
+          [ metavar "FIELDCONTENTS"
+          , help "The contents to insert into the field. Required when creating a new field, optional otherwise"
+          ])
+    <*> optional (option readFieldVisibility $ mconcat
+          [ long "visibility"
+          , short 'V'
+          , metavar "VISIBILITY"
+          , help "Whether to mark this field as 'public' or 'private'"
+          ])
```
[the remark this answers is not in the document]
dcastro: Oh good point, I hadn't considered that. I guess the implementation is also wrong, because the spec says "public" should be the default. I'll change this.

lib/CLI/Parser.hs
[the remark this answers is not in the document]
dcastro: At this point, we have no global arguments, so I didn't include that.

In #19 we'll need a global argument, so at that point we can just wrap `SomeCommand` & the global args in some other data type (which I see you've already done).

DK318 (commented):
Seems good to me

DK318 (approved):

lib/CLI/Types.hs
dcastro: I've rebased on `main`, and changed `coTags` to a set.

lib/CLI/Types.hs
```diff
@@ -0,0 +1,134 @@
...
+module CLI.Types where
+
+import Data.Text (Text)
+import Data.Time.Compat (Day, UTCTime , Year)
+import Data.Time.Calendar.Month.Compat (Month)
+import Entry (FieldKey, FieldVisibility, EntryTag)
+import Coffer.Path (Path, EntryPath)
+import Data.Set (Set)
+
+data Command
```
Heimdell, line 14: Weren't they GADTs? Or is it the state before turning into GATDs?

lib/Coffer/Path.hs
```diff
@@ -0,0 +1,195 @@
...
+
+  where
+    stripPrefix s = fromMaybe s $ T.stripPrefix "/" s
+    stripSuffix s = fromMaybe s $ T.stripSuffix "/" s
+
+-- | An entry's full path (directory + entry's name).
+newtype EntryPath = EntryPath { unEntryPath :: NonEmpty PathSegment }
+  deriving stock (Show, Eq)
+
+makeLensesFor [("unEntryPath", "isoEntryPath")] ''EntryPath
```
Heimdell, line 113: Why `isoEntryPath` and not ~[coerced](https://hackage.haskell.org/package/lens-5.1/docs/Control-Lens-Iso.html#v:coerced)~ `pathSegments`? The `EntryPath` and its segments are entities from domain.

The code style demands it, but I think it would be more readable at use sites as `pathSegments`.

Heimdell (approved):

lib/CLI/Types.hs
dcastro: > Or is it the state before turning into GATDs?

Yup, they were turned into GADTs in the next PR.

lib/Coffer/Path.hs
dcastro: I did it this way mostly because there is no function in `Control.Lens.TH` that (1) lets me choose the generated fields' names _and_ (2) generates a `HasPathSegments` typeclass.

`makeLensesWith` does (2) but not (1). `makeLensesFor` does (1) but not (2)  :disappointed: 

Anyway, I just pushed a fixup where I simply wrote my own `HasPathSegments` typeclass ^^

Heimdell (approved):

PR #31, opened 2022-02-28
[#29] Implement CLI commands

dcastro:
This PR implements most CLI commands.

The only missing command is the `rename` command, which will be implemented in #12

### DIFF
--- a/lib/Backend/Commands.hs
+++ b/lib/Backend/Commands.hs
@@ -1,0 +1,458 @@
+-- SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io>
+--
+-- SPDX-License-Identifier: MPL-2.0
+
+module Backend.Commands where
+
+import Fmt ( pretty )
+import Control.Lens hiding (view)
+import CLI.Types
+import Polysemy
+import Control.Monad.State
+import Data.Time (getCurrentTime, UTCTime, utctDay)
+import qualified Data.Text as T
+import qualified Data.HashMap.Strict as HashMap
+import GHC.Exts (Down(..), sortWith)
+import Data.Maybe (fromMaybe)
+import Data.Time.Calendar.Compat (pattern YearMonthDay)
+import Data.Time.Calendar.Month.Compat (pattern MonthDay)
+import Control.Lens (view)
+import Data.Functor (($>))
+import Polysemy.Error (throw, Error)
+import Control.Monad.Extra (whenM, whenJust)
+import Data.List.NonEmpty (NonEmpty)
+import qualified Data.List.NonEmpty as NE
+import Data.Set (Set)
+import qualified Data.Set as Set
+
+import Backend (BackendEffect, listSecrets, readSecret, writeSecret, deleteSecret)
+import Entry (Entry, Field, FieldKey, value, fields, dateModified, newEntry, newField, visibility, FieldVisibility(..), EntryTag)
+import qualified Entry as E
+import Coffer.Directory (Directory)
+import qualified Coffer.Directory as Dir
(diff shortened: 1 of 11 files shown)


lib/Coffer/Directory.hs
```diff
@@ -0,0 +1,94 @@
...
+
+----------------------------------------------------------------------------
+-- Directory
+----------------------------------------------------------------------------
+
+data Directory = Directory
+  { _entries :: [Entry]
+  , _subdirs :: HashMap PathSegment Directory
+  }
+  deriving stock (Show, Eq)
```
DK318, line 30: Why `Directory` doesn't contain the path to itself?

lib/Coffer/Directory.hs
```diff
@@ -0,0 +1,94 @@
...
+
+emptyDir :: Directory
+emptyDir = Directory mempty mempty
+
+singleton :: Entry -> Directory
+singleton entry = insertEntry entry emptyDir
+
+allEntries :: Directory -> [Entry]
+allEntries (Directory entries subdirs) =
+  entries <> foldMap allEntries subdirs
```
DK318, line 42: It would be good having a function, which lists entries **only** in a specific directory.

lib/Backend/Commands.hs
```diff
@@ -0,0 +1,453 @@
...
+  CmdDeleteField opts -> deleteFieldCmd opts
+  CmdFind opts -> findCmd opts
+  CmdRename opts -> renameCmd opts
+  CmdCopy opts -> catchAndReturn $ copyCmd opts
+  CmdDelete opts -> catchAndReturn $ deleteCmd opts
+  CmdTag opts -> catchAndReturn $ tagCmd opts
+  where
+    catchAndReturn :: forall e r. Sem (Error e ': r) e -> Sem r e
+    catchAndReturn action =
+      either id id <$> runError @e action
```
DK318, line 46: This function could be in some `Util` module.

lib/Backend/Commands.hs
```diff
@@ -0,0 +1,453 @@
...
+    -- | Checks that the paths we're trying to copy entries to aren't already occupied
+    -- by an existing entry.
+    checkDestinatationIsNotEntry :: [CopyOperation] -> Sem r ()
+    checkDestinatationIsNotEntry operations = do
+      invalidOperations <- flip filterM operations \(CopyOperation _ new) -> do
+        pathIsEntry (new ^. E.path)
+      when (not (null invalidOperations)) do
+        invalidOperations
+          <&> (\(CopyOperation old new) -> (old ^. E.path, new ^. E.path) )
+          & throw . CPREntryAlreadyExists
```
DK318, line 325: Looks like copy-paste. Maybe it would be better to create function `checkCopyOperations pred exception`?

lib/Coffer/Directory.hs
dcastro: Because there was no need for that.

You can derive a directory's path by tracing all the `PathSegments` in `_subdirs`.

lib/Coffer/Directory.hs
dcastro: There was no need for that function to implement these CLI commands. We can add that function when a need for it arises.

lib/Backend/Commands.hs
dcastro: Agreed, I've added a `Util` module

lib/Backend/Commands.hs
dcastro: Done

Heimdell (approved):

PR #32, opened 2022-03-04
[#23] Add `bats` tests

### LINKED ISSUE #23: Add a `bats` test suite
We should setup a `bats` test suite for end-to-end tests for the CLI.

--- 

Once we do that, we should also [delete the doctests in `Backend.Vault.Kv.Internal.Routes`](https://github.com/serokell/coffer/blob/c4eacf215089dd40801f1292bda830c3ed18cd75/lib/Backend/Vault/Kv/Internal.hs#L258-L267).

```hs
-- $setup
-- >>> import Network.HTTP.Client.TLS (tlsManagerSettings)
-- >>> import Network.HTTP.Client     (newManager)
-- >>> import Servant.Clieng          (BaseUrl, Https)
--
-- >>> let vaultToken   = "TOKEN"
-- >>> let vaultAddress = "vault.example.org"
--
-- >>>     manager     <- newManager tlsManagerSettings
-- >>> let clientEnv   <- mkClientEnv manager (BaseUrl Https vaultAddress 8200 "")
```

Doctests should not depend on external services, we should be able to run them in isolation without any additional setup. So `bats` is a much better tool for this purpose than doctest.

DK318:
Problem: In #29, we implemented almost every CLI action, but we haven't
written any tests for them yet.

Solution: write golden tests using `bats`.

## Related issue(s)
- Resolves #23 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

### DIFF
--- a/tests/golden/copy-command/copy-command.bats
+++ b/tests/golden/copy-command/copy-command.bats
@@ -1,0 +1,175 @@
+# SPDX-FileCopyrightText: 2020 Serokell <https://serokell.io>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+#!/usr/bin/env bats
+
+load '../helpers/bats-support/load'
+load '../helpers/bats-assert/load'
+load '../helpers'
+
+@test "copy an entry" {
+  coffer create /a
+
+  run coffer copy /a /b
+
+  assert_success
+  assert_output "[SUCCESS] Copied '/a' to '/b'."
+
+  run cleanOutput coffer view /
+  assert_output - <<EOF
+/
+  a - [2000-01-01 01:01:01]
+  b - [2000-01-01 01:01:01]
+EOF
+}
+
+@test "copy a directory" {
+  coffer create /a/b/c
+  coffer create /a/x
+  coffer create /diff/entry
+
+  run coffer copy /a /diff
+
+  assert_success
+  assert_output - <<EOF
+[SUCCESS] Copied '/a/x' to '/diff/x'.
+[SUCCESS] Copied '/a/b/c' to '/diff/b/c'.
+EOF
+
+  run cleanOutput coffer view /
+  assert_output - <<EOF
+/
+  a/
+    x - [2000-01-01 01:01:01]
+    b/
+      c - [2000-01-01 01:01:01]
+  diff/
+    entry - [2000-01-01 01:01:01]
+    x - [2000-01-01 01:01:01]
+    b/
+      c - [2000-01-01 01:01:01]
+EOF
+}
+
+@test "copy one directory to other" {
+  coffer create /a/b/c
+  coffer create /c/d/e
+
+  run coffer copy /a /c
+
+  assert_success
+  assert_output "[SUCCESS] Copied '/a/b/c' to '/c/b/c'."
+
+  run cleanOutput coffer view /
+  assert_output - <<EOF
+/
+  a/
+    b/
+      c - [2000-01-01 01:01:01]
+  c/
+    b/
+      c - [2000-01-01 01:01:01]
+    d/
+      e - [2000-01-01 01:01:01]
+EOF
+}
+
+@test "copying an entry fails when it clashes with an existing directory" {
+  coffer create /a
+  coffer create /b/c/d
+
+  run coffer copy /a /b/c
+
+  assert_failure
+  assert_output - <<EOF
+[ERROR] The following entries cannot be copied because a directory already exists at the destination.
+
+Cannot copy '/a' to '/b/c'.
+EOF
+
+  run coffer copy /a /b/c -f
+
+  assert_failure
+  assert_output - <<EOF
+[ERROR] The following entries cannot be copied because a directory already exists at the destination.
+
+Cannot copy '/a' to '/b/c'.
+EOF
+}
+
+@test "copying a directory fails when any entry clashes with an existing entry" {
+  coffer create /a/b/c
+  coffer create /a/b/x
+  coffer create /a/b/y
+
+  coffer create /d/b/c
+  coffer create /d/b/x
+
+  run coffer copy /a /d
+
+  assert_failure
+  assert_output - <<EOF
+[ERROR] The following entries cannot be copied because an entry already exists at the destination.
+Use '--force' or '-f' to overwrite existing entries.
+
+Cannot copy '/a/b/c' to '/d/b/c'.
+Cannot copy '/a/b/x' to '/d/b/x'.
+EOF
+}
+
+@test "copying a directory succeeds when any entry clashes with an existing entry and --force is used" {
+  coffer create /a/b/c
+  coffer create /a/b/x
+  coffer create /a/b/y
+
+  coffer create /d/b/c
+  coffer create /d/b/x
+
+  run coffer copy /a /d -f
+
+  assert_success
+  assert_output - <<EOF
+[SUCCESS] Copied '/a/b/c' to '/d/b/c'.
+[SUCCESS] Copied '/a/b/x' to '/d/b/x'.
+[SUCCESS] Copied '/a/b/y' to '/d/b/y'.
+EOF
+
+  run cleanOutput coffer view /
+  assert_output - <<EOF
+/
+  a/
+    b/
+      c - [2000-01-01 01:01:01]
+      x - [2000-01-01 01:01:01]
+      y - [2000-01-01 01:01:01]
+  d/
+    b/
+      c - [2000-01-01 01:01:01]
+      x - [2000-01-01 01:01:01]
+      y - [2000-01-01 01:01:01]
+EOF
+}
+
+@test "copying a directory fails when any entry clashes with an existing directory" {
+  coffer create /a/b
+  coffer create /x/b/c
+
+  run coffer copy /a /x
+
+  assert_failure
+  assert_output - <<EOF
+[ERROR] The following entries cannot be copied because a directory already exists at the destination.
+
+Cannot copy '/a/b' to '/x/b'.
+EOF
+
+  run coffer copy /a /x -f
+
+  assert_failure
+  assert_output - <<EOF
+[ERROR] The following entries cannot be copied because a directory already exists at the destination.
+
+Cannot copy '/a/b' to '/x/b'.
+EOF
+}
--- a/tests/golden/delete-command/delete-command.bats
+++ b/tests/golden/delete-command/delete-command.bats
@@ -1,0 +1,63 @@
+# SPDX-FileCopyrightText: 2020 Serokell <https://serokell.io>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+#!/usr/bin/env bats
+
+load '../helpers/bats-support/load'
+load '../helpers/bats-assert/load'
+load '../helpers'
+
+@test "delete an entry" {
+  coffer create /a
+  coffer create /b
+
+  run coffer delete /a
+
+  assert_success
+  assert_output "[SUCCESS] Deleted '/a'."
+
+  run cleanOutput coffer view /
+  assert_output - <<EOF
+/
+  b - [2000-01-01 01:01:01]
+EOF
+}
+
+@test "delete not existing entry" {
+  run coffer delete /a
+
+  assert_failure
+  assert_output "[ERROR] Entry or directory not found at '/a'."
+}
+
+@test "delete directory without recursive" {
+  coffer create /a/b/c
+  coffer create /a/d
+
+  run coffer delete /a/b
+
+  assert_failure
+  assert_output - <<EOF
+[ERROR] The path '/a/b' is a directory.
+Use '--recursive' or '-r' to recursively delete all entries.
+
+EOF
+}
+
+@test "delete directory with recursive" {
+  coffer create /a/b/c
+  coffer create /a/d
+
+  run coffer delete /a/b -r
+
+  assert_success
+  assert_output "[SUCCESS] Deleted '/a/b/c'."
+
+  run cleanOutput coffer view /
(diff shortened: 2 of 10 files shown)


tests/golden/copy-command/copy-command.bats
```diff
@@ -0,0 +1,81 @@
...
+  assert_output "[SUCCESS] Entry created at '/c/d/e'."
+
+  run coffer copy /a /c
+  assert_output "[SUCCESS] Copied '/a/b/c' to '/c/b/c'."
+
+  run cleanOutput coffer view /
+  cat copy-one-directory-to-other.gold | assert_output -
+}
+
+@test "trying to copy an entry to directory" {
```
dcastro, line 43: We should add a similar test that checks that an entry cannot overwrite a directory **even if** `--force` is used.

dcastro, line 43: It's hard to tell what the purpose of these tests is by looking at their names alone.

Test names should be descriptive. May I suggest:
* "trying to copy an entry to directory" -> "copying an entry fails when it clashes with an existing directory"
* "copy a directory without force" -> "copying a directory fails when any entry clashes with an existing entry"
* "copy a directory with force" -> "copying a directory succeeds when any entry clashes with an existing entry and --force is used"

tests/golden/copy-command/copy-command.bats
```diff
@@ -0,0 +1,81 @@
...
+  run coffer create /a/b/c
+  assert_output "[SUCCESS] Entry created at '/a/b/c'."
+
+  run coffer create /d/b/c
+  assert_output "[SUCCESS] Entry created at '/d/b/c'."
+
+  run coffer copy /a /d
+
+  assert_failure
+  cat copy-a-directory-without-force.gold | assert_output --partial -
```
dcastro, line 66: This test tries to copy a directory with one entry, and that entry clashes with an existing entry.

To make this test more robust, we should try to copy a directory with 3 entries, where 2 of them clash and the 3rd doesn't.

This way, we'll also be asserting that:
* coffer reports _all_ clashing entries (and not just the first clash).
* coffer _only_ reports clashing entries. Entries that can be safely copied do not show up in the output.

The same reasoning also applies to similar tests.

tests/golden/copy-command/copy-command.bats
dcastro: Think we should also add a test for "copying a directory fails when any entry clashes with an existing directory"

dcastro (changes requested):

tests/golden/copy-command/copy-a-directory-with-force.gold
```diff
@@ -0,0 +1,7 @@
+/
```
dcastro, line 1: Here's a list of things we should also test:
`view`
  - [ ] test a path pointing to a specific entry
  - [ ] test a path pointing to a specific entry with a field filter

`create`
  - [ ] `create` fails if an entry already exists at that path
  - [ ] `create --force` succeeds if an entry already exists at that path
  - [ ] specifying the same tag multiple times is the same as specifying it once
  - [ ] field contents can be empty
  - [ ] field contents can have multiple lines

`tag`
  - [ ] fails when the path doesn't exist

`set-field`
  - [ ] can update an existing field's contents
  - [ ] fails when the path doesn't exist

`find`
  - [ ] sort by a field's value (asc/desc)
  - [ ] sort by a field's date (asc/desc)
  - [ ] check that `coffer find PATH` excludes entries that are not in `PATH`
  - [ ] check that `coffer find` returns the intersection (and not the union) of all filters
  - [ ] check that `coffer find PATH TEXT` searches for `TEXT` in the entry's path segments (including the entry name) and tags
  - [ ] check that directories where no matches were found do not appear in the result
  - [ ] check that PATH can be an entry's path
  - [ ] check that using a path that doesn't exist returns no matches
  - [ ] test `--filter-field`

`delete-field`
  - [ ] Fails when the path doesn't exist
  - [ ] Fails when the path is a directory

dcastro (changes requested):

dcastro:
Can you please also make a small amendment to the `make bats` target? To make it easier for us to run a specific test (or subset of tests)

```
# Usage:
#   * make bats
#   * make bats FILTER="test name"
bats:
	git submodule update --init --recursive
	./tests/golden/helpers/bats/bin/bats ./tests/golden/** $(if $(FILTER),-f "$(FILTER)",)
```

tests/golden/copy-command/copy-command.bats
```diff
@@ -0,0 +1,81 @@
...
+
+  run coffer copy /a /b
+  assert_output "[SUCCESS] Copied '/a' to '/b'."
+
+  run cleanOutput coffer view /
+  cat copy-an-entry.gold | assert_output -
+}
+
+@test "copy a directory" {
+  run coffer create /a/b/c
```
dcastro, line 19: Try creating 1 more entry in the `/a` directory, and then check that the output of `coffer copy` mentions _all_ copied entries.

Also try creating 1 entry in a different directory, and check that it _doesn't_ appear in the output of `coffer copy`.

dcastro (changes requested):

tests/golden/copy-command/copy-command.bats
```diff
@@ -0,0 +1,81 @@
...
+  assert_output "[SUCCESS] Entry created at '/a/b/c'."
+
+  run coffer copy /a /
+  assert_output "[SUCCESS] Copied '/a/b/c' to '/b/c'."
+
+  run cleanOutput coffer view /
+  cat copy-a-directory.gold | assert_output -
```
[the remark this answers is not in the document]
DK318: Is it okay to write like this?
````bash
...
  EXPECTED=$(cat <<EOF
/
  a - [2000-01-01 01:01:01]
  b - [2000-01-01 01:01:01]
EOF
)
  assert_output "$EXPECTED"
...
````
Because I don't know how to write heredocs in `assert_output` command.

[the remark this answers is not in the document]
Comment (dcastro) [reply tests/golden/copy-command/copy-command.bats:26]: You can add a `-` to tell `assert_output` to read from stdin, so I think something like this should work:

```
  assert_output - <<EOF
/
  a - [2000-01-01 01:01:01]
  b - [2000-01-01 01:01:01]
EOF
```

There's an example [in the docs](https://github.com/bats-core/bats-assert#standard-input-heredocs-and-herestrings), (although the example is using herestrings instead of heredocs).

File: tests/golden/copy-command/copy-command.bats (hunk 1 above)
Comment (DK318) [reply tests/golden/copy-command/copy-command.bats:43]: I'll add this test in #14, because now it fails/

File: tests/golden/copy-command/copy-command.bats (hunk 5 above)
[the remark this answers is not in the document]
Comment (DK318) [reply tests/golden/copy-command/copy-command.bats]: Done

File: tests/golden/view-command/view-command.bats [hunk 6]
```diff
@@ -0,0 +1,36 @@
...
+
+  run cleanOutput coffer view /a/b
+  cat view-at-specified-directory-2.gold | assert_output -
+
+  run cleanOutput coffer view /kek
+  cat view-at-specified-directory-3.gold | assert_output -
+}
+
+@test "view specified field" {
+  run coffer create a/b/c --field test=kek
```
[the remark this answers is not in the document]
DK318: Done

tests/golden/delete-field-command/delete-field-command.bats
```diff
@@ -0,0 +1,29 @@
...
+
+load '../helpers/bats-support/load'
+load '../helpers/bats-assert/load'
+load '../helpers'
+
+@test "delete field" {
+  run coffer create /a/b --field test=test
+  assert_output "[SUCCESS] Entry created at '/a/b'."
+
+  run cleanOutput coffer view /
```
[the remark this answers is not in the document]
DK318: Done

tests/golden/delete-field-command/delete-field-command.bats
```diff
@@ -0,0 +1,29 @@
...
+  cat delete-field-1.gold | assert_output -
+
+  run coffer delete-field /a/b test
```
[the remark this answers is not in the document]
DK318: Done

tests/golden/delete-field-command/delete-field-command.bats
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bats
+
+load '../helpers/bats-support/load'
+load '../helpers/bats-assert/load'
+load '../helpers'
+
+@test "delete field" {
+  run coffer create /a/b --field test=test
+  assert_output "[SUCCESS] Entry created at '/a/b'."
```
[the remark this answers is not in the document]
DK318: Done

tests/golden/copy-command/copy-command.bats
DK318: Done

tests/golden/copy-command/copy-command.bats
DK318: Done

tests/golden/copy-command/copy-command.bats
DK318: Done

tests/golden/copy-command/copy-a-directory-with-force.gold
DK318: Have written all these tests, except for `field contents can have multiple lines` in `create` section. I don't know how to create a field with multiple lines in its contents.

dcastro (changes requested):

tests/golden/copy-command/copy-a-directory-with-force.gold
dcastro: > Have written all these tests, except for field contents can have multiple lines in create section. I don't know how to create a field with multiple lines in its contents.

You can do something like this:
```
coffer create /path --field "user=$(print "first\nsecond" )"
```

tests/golden/copy-command/copy-command.bats
dcastro: It should work here though:

```
$ coffer create /a/b  
$ coffer create /x/b/c
$ coffer copy /a /x   
[ERROR] The following entries cannot be copied because a directory already exists at the destination.

Cannot copy '/a/b' to '/x/b'.
```

The entry `/a/b` clashed with the existing directory `/x/b`.

----

#14 aims to solve a different problem: when copying an entry, and we need to create a directory for that entry to live in, but that directory path is already taken by another entry.

E.g., if we copy an entry to `/a/b/c`, we need to create the directories `/a` and `/a/b`. But if the paths `/a` or `/a/b` are already taken by an existing entry, the operation fails.

tests/golden/copy-command/copy-command.bats
dcastro: Forgot about this bit:

> Also try creating 1 entry in a different directory, and check that it doesn't appear in the output of coffer copy.

dcastro: Oops, no you didn't, I see it now, it's a bit further down.

dcastro: No need to run two `coffer copy` commands in that test though, you can just create all 3 entries at once, and then run `coffer copy` once.

dcastro (changes requested):

tests/golden/copy-command/copy-command.bats
DK318: Done

tests/golden/copy-command/copy-a-directory-with-force.gold
DK318: Done

tests/golden/copy-command/copy-command.bats
DK318: Done

tests/golden/delete-command/delete-command.bats
```diff
@@ -0,0 +1,56 @@
...
+load '../helpers/bats-assert/load'
+load '../helpers'
+
+@test "delete an entry" {
+  coffer create /a
+
+  run coffer delete /a
+
+  assert_success
+  assert_output "[SUCCESS] Deleted '/a'."
```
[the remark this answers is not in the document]
DK318: Done

tests/golden/delete-command/delete-command.bats
```diff
@@ -0,0 +1,56 @@
...
+}
+
+@test "delete directory without recursive" {
+  coffer create /a/b/c
+  coffer create /a/d
+
+  run coffer delete /a/b
+
+  assert_failure
+  assert_output --partial - <<EOF
```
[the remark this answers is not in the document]
DK318: Done

tests/golden/find-command/find-command.bats
```diff
@@ -0,0 +1,297 @@
...
+  assert_output - <<EOF
+/
+  a/
+    b/
+      c - [2000-01-01 01:01:01]
+      d - [2000-01-01 01:01:01]
+EOF
+}
+
+@test "coffer find returns the intersection of all filters" {
```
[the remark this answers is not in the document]
DK318: Done

tests/golden/find-command/find-command.bats
```diff
@@ -0,0 +1,297 @@
...
+
+  assert_success
+  assert_output - <<EOF
+/
+  abcd - [2000-01-01 01:01:01]
+    filtering: kek [2000-01-01 01:01:01]
+EOF
+}
+
+@test "coffer find PATH TEXT searches for TEXT in the entry's path segments (including the entry name) and tags" {
```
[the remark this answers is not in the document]
DK318: Done

dcastro (approved):

PR #34, opened 2022-03-05
[#9] Directory check in `create` command

### LINKED ISSUE #9: `create` command allows to have entries and directories with same name

DK318:
Problem: we can create entries and directories with the same names.
Solution: added subfunction `checkForEntriesInEntryPath` which tries to find an existing entry subpath in given path. Then checked if the given entry path is an existing directory or contains existing entries in subpaths.

## Related issue(s)
- Fixed #9

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -80,6 +80,8 @@ main = do
                 [ "An entry already exists at '" <> build path <> "'."
                 , "Use '--force' or '-f' to overwrite it."
                 ]
+              CRDirectoryAlreadyExists path -> printError $ "A directory already exists at '" <> build path <> "'."
+              CRParentPathContainsEntry path -> printError $ "An entry already exists at '" <> build path <> "'."
 
           SomeCommand cmd@(CmdSetField opts) -> do
             let fieldName = sfoFieldName opts
--- a/lib/Backend/Commands.hs
+++ b/lib/Backend/Commands.hs
@@ -34,6 +34,7 @@ import Coffer.Path (Path, mkPath, mkPathSegment, unPathSegment, entryPathName, E
 import qualified Coffer.Path as Path
 import Error (CofferError(..))
 import Coffer.Util (catchAndReturn)
+import Data.Either (rights)
 
 runCommand
   :: (Member BackendEffect r, Member (Embed IO) r, Member (Error CofferError) r)
@@ -77,7 +78,7 @@ viewCmd (ViewOptions path fieldNameMb) = do
             Nothing -> pure $ VREntryNoFieldMatch (entry ^. E.path) fieldName
 
 createCmd
-  :: (Member BackendEffect r, Member (Embed IO) r, Member (Error CreateResult) r)
+  :: forall r. (Member BackendEffect r, Member (Embed IO) r, Member (Error CreateResult) r)
   => CreateOptions -> Sem r CreateResult
 createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
   nowUtc <- embed getCurrentTime
@@ -91,7 +92,10 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
       & E.fields .~ HashMap.fromList allFields
       & E.tags .~ tags
 
-  -- TODO: check that a directory does not exist at the given path.
+  whenM (pathIsDirectory entryPath) do
+    throw $ CRDirectoryAlreadyExists entryPath
+
+  checkForEntriesInEntryPath entryPath
 
   when (not force) do
     whenM (pathIsEntry entryPath) do
@@ -99,6 +103,28 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
 
   void $ writeSecret entry
   pure $ CRSuccess entry
+
+  where
+    -- | When attempting to create an entry at, e.g., @/a/b/c@, the directories
+    -- @/a@ and @/a/b@ will be implicitly created.
+    --
+    -- This function checks whether those paths are already
+    -- occupied by existing entries and, if so, fails.
+    --
+    -- Note: the root path @/@ cannot possibly be occupied by an entry,
+    -- therefore we skip the check for that path.
+    checkForEntriesInEntryPath :: EntryPath -> Sem r ()
+    checkForEntriesInEntryPath entryPath = do
+      let parentDirsExceptRoot = entryPath
+            & Path.entryPathParentDirs
+            & NE.toList
+            -- Ignore paths that cannot possibly point to an entry
+            <&> Path.pathAsEntryPath
+            & rights
+
+      filterM pathIsEntry parentDirsExceptRoot >>= \case
+        (clashed : _) -> throw $ CRParentPathContainsEntry clashed
+        [] -> pure ()
 
 setFieldCmd
   :: forall r
--- a/lib/Coffer/Path.hs
+++ b/lib/Coffer/Path.hs
@@ -14,6 +14,7 @@ module Coffer.Path
   , mkEntryPath
   , entryPathName
   , entryPathParentDir
+  , entryPathParentDirs
   , appendEntryName
   , pathAsEntryPath
   , entryPathAsPath
@@ -34,6 +35,7 @@ import qualified Data.List as List
 -- $setup
 -- >>> import Fmt (pretty, build)
 -- >>> import Control.Lens
+-- >>> import qualified Data.List.NonEmpty as NE
 -- >>> isRight (Right a) = a
 -- >>> unsafeMkPath = isRight . mkPath
 -- >>> unsafeMkEntryPath = isRight . mkEntryPath
@@ -125,6 +127,16 @@ entryPathName = view $ pathSegments . last1 . to unPathSegment
 entryPathParentDir :: Lens' EntryPath Path
 entryPathParentDir = pathSegments . initNE . from pathSegments
 
+-- | Gets all parents paths of given entry path.
+--
+-- >>> NE.toList (entryPathParentDirs (unsafeMkEntryPath "/dir1/dir2/entry")) <&> build
+-- ["/","/dir1","/dir1/dir2"]
+entryPathParentDirs :: EntryPath -> NonEmpty Path
+entryPathParentDirs entryPath =
+  entryPath ^. pathSegments
+    & NE.init
+    & NE.inits
+    <&> Path
 
 -- | Build an `EntryPath` by append the entry's name to its location path.
 --
--- a/lib/CLI/Types.hs
+++ b/lib/CLI/Types.hs
@@ -44,6 +44,8 @@ data ViewResult
 data CreateResult
   = CRSuccess Entry
   | CREntryAlreadyExists EntryPath
+  | CRDirectoryAlreadyExists EntryPath
+  | CRParentPathContainsEntry EntryPath
 
 data SetFieldResult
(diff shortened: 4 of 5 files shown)


app/Main.hs
```diff
@@ -68,6 +68,7 @@ main = do
             [ "An entry already exists at '" <> build path <> "'."
```
dcastro, line 68: Please rebase this branch on top of `diogo/#23-bats-tests`, and then write tests for this PR.

lib/Backend/Commands.hs
```diff
@@ -91,7 +91,27 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
       & E.fields .~ HashMap.fromList allFields
       & E.tags .~ tags
 
-  -- TODO: check that a directory does not exist at the given path.
+    checkForEntriesInEntryPath :: Member BackendEffect r => EntryPath -> Sem r (Maybe EntryPath)
+    checkForEntriesInEntryPath entryPath =
+      let 
+        Path.EntryPath (start NE.:| rest) = entryPath
```
dcastro, line 97: I think it's fine to use type/constructor names like `EntryPath` and `Path` unqualified. Same for the `:|` operator.

lib/Backend/Commands.hs
```diff
@@ -91,7 +91,27 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
       & E.fields .~ HashMap.fromList allFields
       & E.tags .~ tags
 
-  -- TODO: check that a directory does not exist at the given path.
+    checkForEntriesInEntryPath :: Member BackendEffect r => EntryPath -> Sem r (Maybe EntryPath)
```
dcastro, line 94: Since the `r` type variable in `createCmd` and in `checkForEntriesInEntryPath` are supposed to unify, you could do this:

1. Add a `forall r.` to `createCmd`
2. In `checkForEntriesInEntryPath`, remove the `Member BackendEffect r => ` constraint.

Thanks to [`ScopedTypeVariables`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/scoped_type_variables.html), GHC will know the `r` in both functions is the same, so any constraints applied to `r` in `createCmd` will also be observed in `checkForEntriesInEntryPath`.

This way, we avoid repeating constraints.

lib/Backend/Commands.hs
```diff
@@ -91,7 +91,27 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
...
+        Path.EntryPath (start NE.:| rest) = entryPath
+        startPath = Path.EntryPath (start NE.:| [])
+        restPath = Path.Path rest
+      in process startPath restPath
+        where 
+          process :: Member BackendEffect r => EntryPath -> Path -> Sem r (Maybe EntryPath)
+          process _ (Path.Path []) = pure Nothing
+          process entryPath (Path.Path (x : xs)) = 
+            pathIsEntry entryPath >>= \case
+              True -> (pure . Just) entryPath
```
dcastro, line 106: Instead of returning the entry here, and then using `when isJust + throw` in line 113, we could simplify this by throwing the error here.

lib/Backend/Commands.hs
```diff
@@ -91,7 +91,27 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
       & E.fields .~ HashMap.fromList allFields
       & E.tags .~ tags
 
-  -- TODO: check that a directory does not exist at the given path.
+    checkForEntriesInEntryPath :: Member BackendEffect r => EntryPath -> Sem r (Maybe EntryPath)
+    checkForEntriesInEntryPath entryPath =
+      let 
+        Path.EntryPath (start NE.:| rest) = entryPath
+        startPath = Path.EntryPath (start NE.:| [])
+        restPath = Path.Path rest
```
dcastro, line 99: We could simplify this by separating these two bits of logic: 1. build a list of all the paths we need to check and then 2. iterate through them and check if they already point to an entry.

I think we could add a function like this to the `Path` module (which might even be useful in other situations):

```hs
entryPathParentDirs :: EntryPath -> NonEmpty Path
```

such that (in pseudocode):

```hs
entryPathParentDirs "/a/b/c" == [  "/", "/a", "/a/b" ]
```

Then, here in `checkForEntriesInEntryPath`, all we had to do was iterate through those paths.

dcastro (changes requested):
Looks like my previews comments from the previous PR (#10) haven't been addressed yet, so I'm copying them over to this PR/

app/Main.hs
DK318: Done

lib/Backend/Commands.hs

lib/Backend/Commands.hs

lib/Backend/Commands.hs

lib/Backend/Commands.hs

lib/Coffer/Path.hs
```diff
@@ -125,6 +126,14 @@ entryPathName = view $ pathSegments . last1 . to unPathSegment
 entryPathParentDir :: Lens' EntryPath Path
 entryPathParentDir = pathSegments . initNE . from pathSegments
 
+-- | Gets all parents paths of given entry path.
+entryPathParentDirs :: EntryPath -> NonEmpty Path
+entryPathParentDirs entryPath =
+  entryPath ^. pathSegments
+    & NE.inits
+    & NE.init
+    & NE.fromList
```
dcastro, line 135: `NE.fromList` is [an unsafe function](https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-List-NonEmpty.html#v:fromList):

> Raises an error if given an empty list.

Please avoid using it.

lib/Coffer/Path.hs
```diff
@@ -125,6 +126,14 @@ entryPathName = view $ pathSegments . last1 . to unPathSegment
 entryPathParentDir :: Lens' EntryPath Path
 entryPathParentDir = pathSegments . initNE . from pathSegments
 
+-- | Gets all parents paths of given entry path.
```
dcastro, line 129: Please add some doctests or tasty tests, whichever you prefer.

lib/Backend/Commands.hs
```diff
@@ -91,7 +91,19 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
       & E.fields .~ HashMap.fromList allFields
       & E.tags .~ tags
 
-  -- TODO: check that a directory does not exist at the given path.
+    checkForEntriesInEntryPath :: EntryPath -> Sem r ()
```
dcastro, line 94: Nitpick: I think it would be better to define this helper function in a `where` clause at the bottom.

I think it currently breaks the "flow" when reading this `do`-block from top to bottom.

dcastro, line 94: Let's add some haddocks here explaining what/why we're doing this:

```suggestion
    -- | When attempting to create an entry at, e.g., @/a/b/c@, the directories
    -- @/a@ and @/a/b@ will be implicitly created. 
    -- 
    -- This function checks whether those paths are already
    -- occupied by existing entries and, if so, fails.
    --
    -- Note: the root path @/@ cannot possibly be occupied by an entry, so therefore
    -- we skip the check for that path.
    checkForEntriesInEntryPath :: EntryPath -> Sem r ()
```

lib/Backend/Commands.hs
```diff
@@ -91,7 +91,19 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
...
+    checkForEntriesInEntryPath entryPath = do
+      let eParentDirs = forM ((NE.tail . Path.entryPathParentDirs) entryPath) Path.pathAsEntryPath
+      case eParentDirs of
+        Right parentDirs -> filterM pathIsEntry parentDirs >>= \case
+          (clashed : _) -> throw $ CRParentPathContainsEntry clashed
```
dcastro, line 99: I think we should return and report all clashing paths, not just the first one.

lib/Backend/Commands.hs
```diff
@@ -91,7 +91,19 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
       & E.fields .~ HashMap.fromList allFields
       & E.tags .~ tags
 
-  -- TODO: check that a directory does not exist at the given path.
+    checkForEntriesInEntryPath :: EntryPath -> Sem r ()
+    checkForEntriesInEntryPath entryPath = do
+      let eParentDirs = forM ((NE.tail . Path.entryPathParentDirs) entryPath) Path.pathAsEntryPath
```
dcastro, line 96: Just a suggestion: maybe rename this variable to `parentDirsExceptRoot` to make it explicit that we're checking every parent dir path except the root path?

dcastro, line 96: Suggestion: instead of using `tail`, how about this: we get all parent dirs, and then filter out the root path with `rights`:

```suggestion
    let parentDirsExceptRoot = entryPath
          & Path.entryPathParentDirs 
          & NE.toList
          -- Ignore paths that cannot possibly point to an entry
          <&> Path.pathAsEntryPath
          & rights
```

This way, you don't have to pattern match on `Left/Right` after.

dcastro (changes requested):

dcastro:
Please tick off items from the PR checklist as appropriate.

The `Public contracts` section doesn't really apply at the moment, because we haven't released anything yet, so you can tick those off too.

dcastro:
And add the issue number in the "related issues" section, e.g. `Fixed #9`. When a PR has this in its description, github will automatically close the issues when the PR is merged.

lib/Backend/Commands.hs
DK318: I think in this case if we've found clashed entry, then it would be the only one (all parents would be definitely directories).

lib/Coffer/Path.hs
DK318: Done

lib/Coffer/Path.hs
DK318: Added doctest

lib/Backend/Commands.hs

lib/Backend/Commands.hs

Heimdell (approved):

lib/Backend/Commands.hs
```diff
@@ -100,6 +104,28 @@ createCmd (CreateOptions entryPath _edit force tags fields privateFields) = do
...
+    -- Note: the root path @/@ cannot possibly be occupied by an entry,
+    -- therefore we skip the check for that path.
+    checkForEntriesInEntryPath :: EntryPath -> Sem r ()
+    checkForEntriesInEntryPath entryPath = do
+      let parentDirsExceptRoot = entryPath
+            & Path.entryPathParentDirs
+            & NE.toList
+            -- Ignore paths that cannot possibly point to an entry
+            <&> Path.pathAsEntryPath
+            & rights
```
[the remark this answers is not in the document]
dcastro: `filterM` wouldn't fit here, because we also need to cast from `Path` to `EntryPath`.

Though, `mapMaybe` + `rightToMaybe` would work. But I don't think I would personally prefer that over `fmap` + `rights`, in this instance...

dcastro (approved):

PR #41, opened 2022-03-18
[#40] Added `config.toml` configs in `bats` tests directories

### LINKED ISSUE #40: [BUG] `Bats` test are failing.
# Description

After resolving #15 `bats` tests are failing, because there is no `config.toml` in `tests` directory.

# To Reproduce

1. `stack install`
2. `make bats`

# Expected behavior
`bats` tests are passing.

# Environment

- main [d3e4a4c](https://github.com/serokell/coffer/commit/d3e4a4cd72b9afa925f06486ad275d8761c10a15)

DK318:
Problem: `bats` tests are failing because there is no `config.toml` file.
Solution: added `config.toml` config in each directory.

## Related issue(s)
- Fixes #40 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

### DIFF
--- a/tests/golden/copy-command/config.toml
+++ b/tests/golden/copy-command/config.toml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+main_backend = "vault-local"
+
+[[backend]]
+type = "vault-kv"
+name = "vault-local"
+address = "localhost:8200"
+mount = "secret"
+token = "root"
--- a/tests/golden/create-command/config.toml
+++ b/tests/golden/create-command/config.toml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+main_backend = "vault-local"
+
+[[backend]]
+type = "vault-kv"
+name = "vault-local"
+address = "localhost:8200"
+mount = "secret"
+token = "root"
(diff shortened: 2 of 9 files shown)


tests/golden/copy-command/config.toml
```diff
@@ -0,0 +1,12 @@
...
+# SPDX-License-Identifier: MPL-2.0
+
+main_backend = "vault-local"
+
+[[backend]]
+type = "vault-kv"
+name = "vault-local"
+address = "localhost:8200"
+mount = "secret"
+token = "<vault token>"
```
dcastro, line 12: Hmm these files don't have a token, which means I have to edit all of them in order to run these tests....

I think we should hardcode the token as:

```
token = "root"
```

Then I can just start `vault` with:

```
vault server -dev -dev-root-token-id "root"
```

And `make bats` will work without having to modify any files.

dcastro (changes requested):

tests/golden/copy-command/config.toml
dcastro: Wrote a follow-up issue here: https://github.com/serokell/coffer/issues/42

DK318: I've changed these tokens, but have got interesting behavior. If in main `config.toml` token isn't `root`, then tests would fail. I think that `coffer` depends on main config file and on `config.toml` in the specific test directory and I don't know why.

DK318: Done

dcastro (approved):

dcastro:
LGTM :+1: 

Please squash all fixups and merge, and rebase the other PRs again so we can see their tests passing

PR #45, opened 2022-03-22
[#43] `run-bats-tests` script added.

### LINKED ISSUE #43: Automate spinning up/killing vault instances

Right now, in order to run the `bats` tests, developers have to manually start vault (with `-dev` and `-dev-root-token-id "root"`) and then run `make bats`.

When we write tests for #16, some tests will require having two instances of `vault` running, so the process will require even more steps.

I think we should modify the `make bats` target so that it does all the required setup/teardown automatically (and the logic should probably be moved to a bash script).

The script should do something like this:
* Update the git submodules
* Start two instances of `vault` in dev mode
  * with `-dev-root-token-id "root"`
  * listening on non-default ports to avoid clashing with the developer's own instance (e.g. 8210 and 8220).
* Run `bats`
* Kill both vault instances **regardless of whether the tests succeeded or failed**

# Acceptance criteria

* The `make bats` target automatically spins up and kills the necessary vault instances.
* If a user already has an instance running on the default port 8200, `make bats` does not interfere with that.

DK318:

## Related issue(s)
- Fixes #43

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

### DIFF
--- a/scripts/run-bats-tests.sh
+++ b/scripts/run-bats-tests.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+# This script updates bats git submodules, runs two vault instances,
+# then runs bats tests and after all kills vault instances.
+
+git submodule update --init --recursive
+
+FILTER=$1
+
+vault server -dev -dev-root-token-id="root" -dev-listen-address="localhost:8209" > /dev/null 2>&1 &
+vault1_pid=$!
+
+vault server -dev -dev-root-token-id="second" -dev-listen-address="localhost:8210" > /dev/null 2>&1 &
+vault2_pid=$!
+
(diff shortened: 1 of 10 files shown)


scripts/run-bats-tests.sh
```diff
@@ -0,0 +1,27 @@
...
+# SPDX-License-Identifier: MPL-2.0
+
+# This script updates bats git submodules, runs two vault instances,
+# then runs bats tests and after all kills vault instances.
+
+git submodule update --init --recursive
+
+FILTER=$1
+
+vault server -dev -dev-root-token-id="root" > /dev/null 2>&1 &
```
dcastro, line 14: We should start this instance on a non-default port too.

Otherwise, if we use the default port and the developer already has an instance running on that port, the tests will delete all of the developer's entries.

dcastro (changes requested):

scripts/run-bats-tests.sh
DK318: Done

dcastro (approved):

PR #50, opened 2022-03-24
[#48] Appending reset ANSI control sequences char after printing field value

### LINKED ISSUE #48: Reset ANSI control sequences after printing

The contents of a field may contain any and all unicode characters.

This means they can also include [ANSI control sequences](https://en.wikipedia.org/wiki/ANSI_escape_code) to, for example, turn the background red (`\x1b[41;1m`):

```
$ coffer create /dir/entry1 --field user="$(echo "\x1b[41;1mdiogo")"
$ coffer view /
```

image

This means that one field's control sequences can potentially affect the rest of `coffer`'s output.
```
$ coffer create /dir/entry1 --field user="$(echo "\x1b[41;1mdiogo")"
$ coffer create /dir/entry2 --field user="leonid"                   
$ coffer view /    
```

image

To prevent this from happening, we should reset all ANSI control sequences (with `\x1b[0m`) after printing a field's contents.

We could use the [`ansi-terminal`](https://hackage.haskell.org/package/ansi-terminal) package to do this.

Because we're printing the field's contents in multiple different places, I think it would make sense to wrap the `Text` in a newtype, and write a `Buildable` instance that automatically appends the reset sequence.

# Acceptance criteria

* ANSI control sequences are reset after a field's contents is printed on the terminal.

DK318:

## Related issue(s)

- Fixes #48

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

### DIFF
--- a/lib/Backend/Commands.hs
+++ b/lib/Backend/Commands.hs
@@ -24,7 +24,7 @@ import Data.Set (Set)
 import qualified Data.Set as Set
 
 import Backend (BackendEffect, listSecrets, readSecret, writeSecret, deleteSecret, SomeBackend)
-import Entry (Entry, Field, FieldKey, value, fields, dateModified, newEntry, newField, visibility, FieldVisibility(..), EntryTag, path)
+import Entry (Entry, Field, FieldKey, value, fields, dateModified, newEntry, newField, visibility, FieldVisibility(..), EntryTag, path, fieldValue)
 import qualified Entry as E
 import Coffer.Directory (Directory)
 import qualified Coffer.Directory as Dir
@@ -187,7 +187,7 @@ findCmd config (FindOptions qPathMb textMb sortMb filters filterFields) = do
         Nothing -> False
         Just field ->
           case filter of
-            FilterFieldByValue substr -> substr `T.isInfixOf` (field ^. value)
+            FilterFieldByValue substr -> substr `T.isInfixOf` (field ^. value . fieldValue)
             FilterFieldByDate op date -> matchDate op date (field ^. dateModified)
 
   let path = maybe mempty qpPath qPathMb
@@ -231,6 +231,7 @@ findCmd config (FindOptions qPathMb textMb sortMb filters filterFields) = do
       . each
       . filtered (\field -> field ^. visibility == Private)
       . value
+      . fieldValue
       .~ "[private]"
 
     matchDate :: FilterOp -> FilterDate -> UTCTime -> Bool
--- a/lib/Entry.hs
+++ b/lib/Entry.hs
@@ -11,6 +11,8 @@ module Entry
   , newFieldKey, newEntryTag, getEntryTag
   , visibility, value, tags, EntryTag
   , FieldVisibility(..)
+  , FieldValue (..)
+  , fieldValue
   )
 where
 
@@ -24,6 +26,8 @@ import qualified Data.Aeson as A
 import Data.Hashable (Hashable)
 import Data.Time (UTCTime)
 import Coffer.Path (EntryPath)
+import System.Console.ANSI (setSGRCode, SGR (Reset))
+import System.Console.ANSI.Codes (csi)
 
 newtype FieldKey = UnsafeFieldKey T.Text
   deriving stock (Show, Eq)
@@ -76,16 +80,31 @@ instance A.FromJSON FieldVisibility where
     "private" -> pure Private
     other -> fail $ "expecting either 'public' or 'private', but found: '" <> T.unpack other <> "'"
 
+newtype FieldValue = FieldValue { unFieldValue :: T.Text }
+  deriving stock (Show, Eq, Ord)
+makeLensesFor [("unFieldValue", "fieldValue")] ''FieldValue
+
+-- | User can use ANSI control sequences in field values.
+-- If some ANSI control sequence contain in field value then we append @reset@ ANSI control sequence.
+-- Otherwise, we just return wrapped text.
+-- You can see explanation here (https://github.com/serokell/coffer/issues/48)
+instance Buildable FieldValue where
+  build (FieldValue t) =
+    if T.pack (csi [] "") `T.isInfixOf` t then
+      build t <> build (setSGRCode [Reset])
+    else
+      build t
+
 data Field =
   Field
   { fDateModified :: UTCTime
   , fVisibility :: FieldVisibility
-  , fValue :: T.Text
+  , fValue :: FieldValue
   }
   deriving stock (Show, Eq)
 makeLensesWith abbreviatedFields ''Field
 
-newField :: UTCTime -> T.Text -> Field
+newField :: UTCTime -> FieldValue -> Field
 newField time value =
   Field
   { fDateModified = time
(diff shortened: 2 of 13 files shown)


tests/golden/find-command/find-command.bats
```diff
@@ -18,8 +18,8 @@ load '../helpers'
 /
   secrets/
     google - [2000-01-01 01:01:01]
-      email: johndoe@gmail.com [2000-01-01 01:01:01]
-      pw: [private]            [2000-01-01 01:01:01]
+      email: johndoe@gmail.com$reset [2000-01-01 01:01:01]
+      pw: [private]$reset            [2000-01-01 01:01:01]
```
dcastro, line 22: Hmm, maybe appending the reset sequence to _all_ field is a bit too cumbersome, it affects the output of all tests even where there's no need to.

What if we try to make the `Buildable` instance a bit smarter? E.g. append the reset sequence _only when_ the string contains `\x1b[` (the sequence for [Control Sequence Introducers](https://en.wikipedia.org/wiki/ANSI_escape_code#CSIsection))

lib/Entry.hs
```diff
@@ -76,11 +78,18 @@ instance A.FromJSON FieldVisibility where
     "private" -> pure Private
     other -> fail $ "expecting either 'public' or 'private', but found: '" <> T.unpack other <> "'"
 
+newtype FieldValue = FieldValue { fvValue :: T.Text }
+  deriving stock (Show, Eq)
+makeLensesWith abbreviatedFields ''FieldValue
+
+instance Buildable FieldValue where
+  build (FieldValue t) = build t <> build (setSGRCode [Reset])
```
dcastro, line 86: Please add an haddock comment to the `Buildable` instance explaining why we're doing this. A link to the #48 issue might also be helpful.

lib/Entry.hs
```diff
@@ -76,11 +78,18 @@ instance A.FromJSON FieldVisibility where
     "private" -> pure Private
     other -> fail $ "expecting either 'public' or 'private', but found: '" <> T.unpack other <> "'"
 
+newtype FieldValue = FieldValue { fvValue :: T.Text }
```
dcastro, line 81: We should now be using this type everywhere (e.g. see `CLI.Types`, possibly other modules too).

dcastro (changes requested):

tests/golden/find-command/find-command.bats
DK318: Done

lib/Entry.hs

lib/Entry.hs
DK318: Done (I think)

lib/Entry.hs
dcastro: I see the new comment describes what we're doing, but it doesn't say _why_ (e.g., what issues are we trying to prevent by appending `reset`?).

I think adding a link to #48 would be sufficient - future readers can just go and read the issue to understand why we did this.

lib/Entry.hs
dcastro: See also:
* `newEntry` should take a `FieldValue`
* `parseFieldContentsEof` should return a `FieldValue`
*

lib/Entry.hs
```diff
@@ -76,11 +79,25 @@ instance A.FromJSON FieldVisibility where
...
```
dcastro, line 82: According to the [style guide](https://github.com/serokell/coffer/blob/main/docs/code-style.md#records), a newtype's field accessor should be named `un<newtype name>` (in this case, `unFieldValue`)

lib/Backend/Commands.hs
```diff
@@ -227,7 +227,7 @@ findCmd (FindOptions pathMb textMb sortMb filters filterFields) = do
             Just (SortByEntryName, direction) ->
               sortWith' direction (view $ E.path . to entryPathName) entries
             Just (SortByFieldValue fieldName, direction) -> do
-              let getFieldValue e = e ^? fields . ix fieldName . value
+              let getFieldValue e = e ^? fields . ix fieldName . value . value
```
dcastro, line 230: There's no need to unwrap the `FieldValue` to `Text` here.

tests/golden/helpers.bash
```diff
@@ -22,3 +22,6 @@ setup () {
 cleanOutput () {
   $@ | sed -r 's/\[[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2}\]$/[2000-01-01 01:01:01]/g'
 }
+
+# Reset ANSI control sequences char.
+reset=$(printf '\x1b[0m')
```
dcastro, line 27: I think we now need tests specific for this use PR. We want to check that all commands that output a "field contents" append the `reset` sequence when necessary.

dcastro (changes requested):

lib/Entry.hs
DK318: Don't get where `newEntry` should take a `FieldValue`. At this moment its signature is
````haskell
newEntry :: EntryPath -> UTCTime -> Entry
````

File: lib/Entry.hs (hunk 4 above)
Comment (DK318) [reply lib/Entry.hs:82]: After that fix, the `value` lens won't be generated.

File: lib/Entry.hs (hunk 2 above)

File: tests/golden/helpers.bash (hunk 6 above)
Comment (DK318) [reply tests/golden/helpers.bash:27]: Done

File: lib/Entry.hs (hunk 3 above)
Comment (dcastro) [reply lib/Entry.hs:82]: Oops I meant `newField` :sweat_smile:

File: lib/Entry.hs (hunk 4 above)
Comment (dcastro) [reply lib/Entry.hs:82]: It will generate a `fieldValue` lens instead.

If you add this at the top of the module, you can see the code generated by TH:

```hs
{-# OPTIONS_GHC -ddump-splices -dsuppress-uniques #-}

newtype FieldValue = FieldValue { unFieldValue :: T.Text }
  deriving stock (Show, Eq, Ord)
makeLensesWith abbreviatedFields ''FieldValue
```

```
/home/dc/dev/serokell/coffer/lib/Entry.hs:87:1-45: Splicing declarations
    makeLensesWith abbreviatedFields ''FieldValue
  ======>
    class HasFieldValue s a | s -> a where
      fieldValue :: Lens' s a
    instance HasFieldValue FieldValue Text where
      {-# INLINE fieldValue #-}
      fieldValue = (iso (\ (FieldValue x) -> x)) FieldValue
```


Even better, since this is a newtype, we could use `makeLensesFor` to generate an `Iso` instead of a `Lens`.

```hs
makeLensesFor [("unFieldValue", "fieldValue")] ''FieldValue
```
```
/home/dc/dev/serokell/coffer/lib/Entry.hs:88:1-59: Splicing declarations
    makeLensesFor [("unFieldValue", "fieldValue")] ''FieldValue
  ======>
    fieldValue :: Iso' FieldValue Text
    fieldValue = (iso (\ (FieldValue x) -> x)) FieldValue
    {-# INLINE fieldValue #-}
```

lib/Entry.hs

lib/Entry.hs

dcastro (approved):

MagicRB:
I would compeletely disallow ANSI sequences in field values but this looks ok i guess, only danger is that an attacker can aribtrarily affect the user's terminal, but the code itself LGTM

PR #54, opened 2022-04-01
[#25] Creating one connection manager for all `BackendEffect` actions

### LINKED ISSUE #25: Avoid creating many connection managers
In `runVaultIO`, we're creating a new connection manager on every `BackendEffect` action:

```hs
    env <-
      case url of
        (BaseUrl Http _ _ _) -> do
          manager <- embed $ newManager defaultManagerSettings
          pure $ mkClientEnv manager url
        (BaseUrl Https _ _ _) -> do
          manager <- embed $ newManager tlsManagerSettings
          pure $ mkClientEnv manager url
```

This can be observed by adding a trace statement:
```hs
          manager <- embed $ do
            traceM "Creating manager"
            newManager defaultManagerSettings
```

And then running `coffer view /` (after creating a few entries):

```
$ cabal run exe:coffer -- view /

"Creating manager"
"Creating manager"
"Creating manager"
"Creating manager"
"Creating manager"
"Creating manager"
"Creating manager"
"Creating manager"
/
  test/
    ...
```

We should refactor this so only 1 connection manager is created.

DK318:

## Related issue(s)
- Fixes #25

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
  - [ ] Any modifications of public contracts comply with the [Evolution
  - [ ] I added an entry to the [changelog](../tree/master/CHANGES.md) if my changes are visible to the users
        and
  - [ ] provided a migration guide for breaking changes if possible


lib/Backend/Vault/Kv.hs
```diff
@@ -87,14 +88,24 @@ data CofferSpecials =
...
 getEnv backend =
   case url of
     (BaseUrl Http _ _ _) -> do
-      manager <- embed $ newManager defaultManagerSettings
+      manager <- getManager defaultManagerSettings
       pure $ mkClientEnv manager url
     (BaseUrl Https _ _ _) -> do
-      manager <- embed $ newManager tlsManagerSettings
+      manager <- getManager tlsManagerSettings
       pure $ mkClientEnv manager url
```
dcastro, line 109: Say we have 2 backends in the config, `b1` uses HTTP and `b2` uses HTTPS.

If we make a request to `b1` first, `getEnv` and `getManager` will create a manager with `defaultManagerSettings`.
If we then make a request to `b2`, `getEnv` and `getManager` will reuse the previous manager, and fail with `TlsNotSupported` (I think).

We need to have (at most) 2 managers, one for HTTP and another for HTTPS requests.

lib/Backend.hs
```diff
@@ -20,8 +20,10 @@ import Polysemy.Error                (Error)
 import Polysemy
 import Coffer.Path (EntryPath, Path)
 import BackendName (BackendName)
+import Network.HTTP.Client (Manager)
+import Polysemy.State (State)
 
-type Effects r = (Member (Embed IO) r, Member (Error CofferError) r)
+type Effects r = (Member (Embed IO) r, Member (State (Maybe Manager)) r, Member (Error CofferError) r)
```
dcastro, line 26: Using `State` will work here for now, but `State` is meant for single-threaded applications and we'll get into trouble when we introduce parallelism in #4.

I think it's simpler to just create two `Manager`s when the app starts.

dcastro (changes requested):

lib/Backend.hs
DK318: What about using `Reader` then?

lib/Backend/Vault/Kv.hs

lib/Backend.hs
dcastro: > What about using Reader instead of State?

Yup, sounds good to me!

lib/Backend.hs
```diff
@@ -20,8 +20,10 @@ import Polysemy.Error                (Error)
 import Polysemy
 import Coffer.Path (EntryPath, Path)
 import BackendName (BackendName)
+import Network.HTTP.Client (Manager)
+import Polysemy.Reader (Reader)
 
-type Effects r = (Member (Embed IO) r, Member (Error CofferError) r)
+type Effects r = (Member (Embed IO) r, Member (Reader (Manager, Manager)) r, Member (Error CofferError) r)
```
dcastro, line 26: I would create a proper `ConnectionManagers` record with named fields, instead of a tuple. Otherwise, it's easy to forget which one is the HTTP manager and which one is the HTTPS manager.

DK318: Done

dcastro (approved):

dcastro:
> having to create a tls connection manager for every backend is less then ideal

How so?

> What we need is a way for each backend to store some data as `a` somehow

I didn't really understand what you meant here. What do you mean? How is this related to connection managers?

MagicRB:
My new pass backend doesnt need a tls connection mamager, so adding a `Reader Tls...` as a dependency for all backends is not ideal https://github.com/serokell/coffer/pull/54/files#diff-a0dd0bee522a338924cfebf12a6c79b6cb8a9bfdd093e249d370306060b45480R32

MagicRB:
We're once again making the frontend take responsibility for thing which should not be its responsibility, also such state will make it hard for #56 to work.

dcastro:
@MagicRB Oh I see, I hadn't considered that some backends might not need a connection manager at all.

Yup, that makes sense. Would you please open a PR that solves the issue in #16?

MagicRB:
I think I've come up with a way to do this, I'll  get it done today

PR #58, opened 2022-04-04
[#52] Show qualified paths in `coffer` messages

### LINKED ISSUE #52: Qualified paths in `coffer` messages
In #18 I've supported qualified paths (we can specify backend which certain path is related to) but now both success and error messages show only path without backend
````
coffer copy 'b1#/a' 'b2#/a'  
[SUCCESS] Copied '/a/b' to '/a/b'.
````
And it should look like
````
coffer copy 'b1#/a' 'b2#/a'  
[SUCCESS] Copied 'b1#/a/b' to 'b2#/a/b'.
````
This issue is blocked by #18 and #21.

# Acceptance criteria
All `coffer` messages show qualified paths.

### PR DESCRIPTION
**Pull request by DK318**

## Related issue(s)
- Fixes #52 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -11,14 +11,13 @@ import CLI.Parser
 import CLI.PrettyPrint
 import CLI.Types
 import Coffer.Directory qualified as Dir
-import Coffer.Path (EntryPath, Path)
+import Coffer.Path (EntryPath, Path, QualifiedPath(qpPath))
 import Config (Config(..), configCodec)
 import Control.Lens
 import Control.Monad (forM, forM_, when)
 import Data.Maybe (fromMaybe)
 import Data.Text (pack)
 import Data.Text.IO qualified as TIO
-import Entry (Entry, path)
 import Entry qualified as E
 import Error
 import Fmt
@@ -95,12 +94,14 @@ main = do
             [ "The entry at '" +| path |+ "' does not yet have a field '" +| fieldName |+ "'."
             , "In order to create a new field, please include the 'FIELDCONTENTS' argument."
             ]
-          SFRSuccess entry -> do
+          SFRSuccess qEntry -> do
+            let entry = qpPath qEntry
+            let qPath = view E.path <$> qEntry
             let field = entry ^?! E.fields . ix fieldName
             printSuccess $
               "Set field '" +| fieldName |+
               "' (" +| (field ^. E.visibility) |+
-              ") at '" +| entry ^. E.path |+
+              ") at '" +| qPath |+
               "' to:\n" +| (field ^. E.value) |+ ""
 
       SomeCommand cmd@(CmdDeleteField opts) -> do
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: lib/Backend/Commands.hs [hunk 1]
```diff
@@ -257,35 +263,38 @@ renameCmd
...
-    forM_ pathsToDelete \(CopyOperation old _) -> do
+    forM_ pathsToDelete \(CopyOperation old _ _ _) -> do
       let qPath = QualifiedPath oldBackendNameMb (Path.entryPathAsPath (old ^. path))
       void $ catchAndReturn $ deleteCmd config (DeleteOptions dryRun qPath False)
 
   pure $ CPRSuccess $ getOperationPaths <$> operations
 
 data CopyOperation = CopyOperation
   { coOld :: Entry
+  , coOldBackendNameMb :: Maybe BackendName
```
dcastro, line 289: I wonder if some of the code could be simplified if we stored a `coOld :: QualifiedPath Entry` instead of `Entry + Maybe BackendName`....

Then, in `getOperationPaths`, to transform a `QualifiedPath Entry` into a `QualifiedPath EntryPath` we could:
* Derive `Functor` for `QualifiedPath`
* And do `view E.path <$> old`

We could derive `Functor` for `QualifiedPath`,

dcastro (changes requested):

app/Main.hs
```diff
@@ -22,7 +22,7 @@ import Backend.Interpreter
 import Error
```
dcastro, line 22: The `set-field` command still shows unqualified paths, see line 110 of `tests/golden/set-field-command/set-field-command.bats`

Maybe we should change `SFRSuccess Entry` to `SFRSuccess (QualifiedPath Entry)`?

dcastro: In general, I think most of the `*Result` constructors (`SetFieldResult`, `CreateResult`, `CreateError`, etc) that take an `Entry` should take a `QualifiedPath Entry` instead.

Here's another one that's missing:

```
$ coffer -c tests/golden/tag-command/config.toml create 'vault-local#/a'
$ coffer -c tests/golden/tag-command/config.toml create 'second#/a'     
$ coffer -c tests/golden/tag-command/config.toml copy 'vault-local#/a' 'second#/a'  
[ERROR] The following entries cannot be copied:

'vault-local#/a' to '/a':
  An entry already exists at '/a'.
  Use '--force' or '-f' to overwrite existing entries.
```

The 3rd and 4th lines should show `second#/a` instead of `/a`.

dcastro (approved):

dcastro (changes requested):
Oops accidentally approved, didn't mean to do that.

app/Main.hs
dcastro: Maybe some constructors could be simplified to take only a `QualifiedPath EntryPath` instead of an entire `QualifiedPath Entry`?

lib/Backend/Commands.hs
```diff
@@ -111,21 +113,23 @@ setFieldCmd
 setFieldCmd config (SetFieldOptions (QualifiedPath backendNameMb entryPath) fieldName fieldContentsMb visibilityMb) = do
```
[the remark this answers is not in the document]

lib/Backend/Commands.hs
```diff
@@ -111,21 +113,23 @@ setFieldCmd
...
+    updateOrInsert backendNameMb nowUtc entry =
       entry
         & dateModified .~ nowUtc
-        & fields . at fieldName %%~ updateOrInsertField nowUtc
+        & fields . at fieldName %%~ updateOrInsertField backendNameMb nowUtc
 
-    updateOrInsertField :: UTCTime -> Maybe Field -> Sem r (Maybe Field)
-    updateOrInsertField nowUtc = \case
+    updateOrInsertField :: Maybe BackendName -> UTCTime -> Maybe Field -> Sem r (Maybe Field)
+    updateOrInsertField backendNameMb nowUtc = \case
```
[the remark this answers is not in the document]

lib/Backend/Commands.hs

app/Main.hs
DK318: Done

dcastro:
I'm starting to think `Entry` should have a `QualifiedPath EntryPath` instead of an `EntryPath`.

But this is outside the scope of this PR. I'll have a deeper look at this later and re-valuated where and how we should use `QualifiedPath`.

LGTM :+1:

dcastro (approved):

MagicRB:
Regarding the code itself, LGTM. Architecturally:
1. `QualifiedPath` is a weird a name, especially when specialized as `QualifiedPath Entry`. Imo we should rename it and maaybe add a class constraint just so the types are a bit tighter and represent more what we meant
2. I don't think `Entry` should take on a `QualifiedPath EntryPath` or such field. Two entries from two backends should be equivalent, once an entry is read, its not bound to any one backend, it just exists. Adding more data on top of it with `QualifiedPath` or similar is imo the right way to go

MagicRB (approved):

PR #63, opened 2022-04-05
[#39] String interpolation in `coffer` output messages

### LINKED ISSUE #39: Use `nyan-interpolation`

In `Main.hs`, we construct user-facing messages such as:

```hs
            printSuccess $
              "Set field '" +| fieldName |+
              "' to '" +| (field ^. E.value) |+
              "' (" +| (field ^. E.visibility) |+
              ") at '" +| entry ^. E.path |+ "'."
```

This would look less convoluted if we used some kind of interpolation.
Interpolation would also make the code less prone to errors such as forgetting to open/close a single quote `'`.

We recently released [`nyan-interpolation`](https://hackage.haskell.org/package/nyan-interpolation), let's use it here.

I haven't checked if these quasiquotes produced by `nyan-interpolation` could be embedded in `Builder` expressions (I suspect they can). If so, we should also investigate whether we could use it to simplify `CLI.PrettyPrint`.

# Acceptance criteria

We've used interpolation where possible.

DK318:

## Related issue(s)
- Fixes #39

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and


app/Main.hs
```diff
@@ -208,6 +226,6 @@ main = do
       buildErrorMessages errors = do
         forM errors \(from, err) -> do
           let entryTo = getEntryFromCreateError err
-          let header = "'" +| from |+ "' to '" +| entryTo ^. path |+ "':"
+          let header = [int|s|'#{from}' to '#{entryTo ^. path}':|]
           let errorMsg = createErrorToBuilder err
           pure $ unlinesF @_ @Builder $ header : [indentF 2 errorMsg]
```
dcastro, line 231: I didn't test it, but would this work?

```hs
          let entryTo = getEntryFromCreateError err
          pure [int||
            '#{from}' to '#{entryTo ^. path}':
              #{createErrorToBuilder err}
          |]
```

dcastro: Actually, also just noticed that `buildErrorMessages` doesn't have to return `Sem r [Builder]`, it could just return a `[Builder]`.

dcastro (commented):
LGTM, just one small suggestion! :+1:

app/Main.hs
Martoon-00: Oh, I'm afraid no. Just a couple of days ago I witnessed the need for such functionality too, so let the ticket be https://github.com/serokell/nyan-interpolation/issues/11.

dcastro: Ah right, and one of the messages returned by `createErrorToBuilder` does have multiple lines :/

Ok, let's leave it like this for now. @DK318, please create a separate issue to improve this when https://github.com/serokell/nyan-interpolation/issues/11 is done + released.

dcastro: Note: while we're here, we could still change `Sem r [Builder]` -> `[Builder]` in this PR.

sancho20021: @DK318 I checked, `Sem r [Builder]` can be changed to `[Builder]` (as mentioned above). For example, like this:
```
buildErrorMessages = fmap \(from, err) ->
  let entryPath = getEntryFromCreateError err
      header    = [int|s|'#{from}' to '#{entryTo ^. path}':|]
      errorMsg  = createErrorToBuilder err
  in unlinesF @_ @Builder $ header : [indentF 2 errorMsg]
```

sancho20021 (changes requested):
Please, resolve conflicts with the main branch and maybe change the `buildErrorMessages` type :)

Everything else looks good

DK318:
Oh, this PR is really outdated. Still waiting for a review :smile:

dcastro:
> Oh, this PR is really outdated. Still waiting for a review :smile: 

Looks like you have 2 reviews now, I really only had that 1 small comment ^^

DK318:
Oh. This `Review required` got me distracted

app/Main.hs
```diff
@@ -7,26 +7,28 @@ module Main where
...
+import Control.Monad (forM_, when)
+import Data.Text (pack)
+import Data.Text.IO qualified as TIO
+import Fmt
+import Options.Applicative (execParser)
+import Polysemy
+import Polysemy.Error (Error, errorToIOFinal)
+import System.Exit (die, exitFailure)
+import Toml qualified
+
```
dcastro, line 20: Extra newline here

.hlint.yaml
```diff
@@ -6,7 +6,7 @@
 # Settings
```
dcastro, line 6: Since things have changed in `main` since the last rebase, I think we could use interpolation here as well:

* `Buildable` instance for `InternalCommandsError` and `CofferError`
* `Buildable` instance for `VaultError`
* `CLI.Parser.readSum`, `readFieldInfo`, `readSort`, `readFilter`, `readQualifiedPath`, `configPathParser`

app/Main.hs
DK318: Removed

.hlint.yaml
DK318: Done

dcastro (approved):

PR #68, opened 2022-04-10
Diogo/#28 code style

### LINKED ISSUE #28: Make code conform to Serokell style policies.
The code style (including points that `stylish-haskell` cannot fix by itself) does not conform to https://github.com/serokell/style/blob/master/haskell.md.

We should apply the code style referenced above with enabling `BlockArguments`, `LambdaCase` and `ImportQualfiedPost` inside our copy of style configuration file.

dcastro:
## Description

This PR refactors a few small things in the codebase and applies some of the guidelines outlined in `code_style`.

Since this PR touches many files, it's best to review it by reviewing one commit at a time (rather than everything at once).

## Related issue(s)

Fixes #28 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
  - [ ] Any modifications of public contracts comply with the [Evolution
  - [ ] I added an entry to the [changelog](../tree/master/CHANGES.md) if my changes are visible to the users
        and
  - [ ] provided a migration guide for breaking changes if possible


lib/Entry/Json.hs
```diff
@@ -11,20 +11,21 @@ import Data.Aeson qualified as A
...
 
-fieldConverter :: (Prism A.Value A.Value E.Field E.Field)
+fieldConverter :: (Prism A.Value A.Value Field Field)
 fieldConverter = prism' to from
   where
-    to :: E.Field -> A.Value
+    to :: Field -> A.Value
     to field =
       A.object
       [ "date_modified" A..= (field ^. E.dateModified)
```
DK318, line 31: Maybe we can do the same thing with `A.Value` and `A..=` (for e.g. hide `.=` from `Control.Lens`)

lib/Backend/Commands.hs
```diff
@@ -87,9 +87,16 @@ viewCmd config (ViewOptions (QualifiedPath backendNameMb path) fieldNameMb) = do
 
 createCmd
   :: forall r
-   . (Member BackendEffect r, Member (Error CofferError) r, Member (Embed IO) r, Member (Error CreateResult) r)
+   . ( Member BackendEffect r
+     , Member (Error CofferError) r
+     , Member (Embed IO) r
+     , Member (Error CreateResult) r
+     )
```
DK318, line 94: I think we can use [Members](https://hackage.haskell.org/package/polysemy-1.7.1.0/docs/Polysemy.html#t:Members) instead of `Member` here (and in some other places)

DK318 (changes requested):

lib/Entry/Json.hs
dcastro: Done, did the same with `Toml..=`

lib/Backend/Commands.hs
dcastro: Good catch! Done

DK318 (approved):
LGTM

PR #70, opened 2022-04-14
[#66] Rename command optimization

### LINKED ISSUE #66: Optimize deletion of entries in `renameCmd`

In the [implementation of `renameCmd`](https://github.com/serokell/coffer/blob/0296786c0c70583d2d768d5c17cb909845e18e9f/lib/Backend/Commands.hs#L261-L280), we use `deleteCmd` to delete any old entries.

However, reusing `deleteCmd` means we have to make an otherwise unnecessary call to `getEntryOrDirThrow`.

It would be more efficient to directly call `deleteSecret` instead of `deleteCmd`.

# Acceptance criteria

`renameCmd` does not use `deleteCmd`. Instead, it calls `deleteSecret` on every entry that needs to be deleted.

DK318:
## Problem 
At this moment `rename` command reuses `delete` command,
so we do some unnecessary actions like `getEntryOrDirThrow`.

## Solution 
Replaced `deleteCmd` call with `deleteSecret`.

## Related issue(s)

- Fixes #66

Fixed #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/lib/Backend/Commands.hs
+++ b/lib/Backend/Commands.hs
@@ -296,8 +296,7 @@ renameCmd
   -- then we can delete old directory/entry without delete errors.
   unless dryRun do
     forM_ pathsToDelete \(CopyOperation old _) -> do
-      let qPath = QualifiedPath oldBackendNameMb (Path.entryPathAsPath (qpPath old ^. path))
-      void $ catchAndReturn $ deleteCmd config (DeleteOptions dryRun qPath False)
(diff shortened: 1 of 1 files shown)


dcastro (approved):

MagicRB (approved):
Seems fine to me, just calling the effect directly and not going through the command interface. If I understood it right LGTM

PR #73, opened 2022-04-19
[#72] Backend-agnostic `CofferError` GADT

### LINKED ISSUE #72: Create backend-specific error types

As of #60, we're using the following definition for `CofferError`:

```hs
data CofferError
  = ServantError ClientError
  | BackendNotFound BackendName
  | OtherError Builder
```

There are 2 issues here:

1. In the `Backend.Vault.Kv` module, we're constructing `OtherError` with user-facing error messages as strings. We should follow the same pattern we're using everywhere else and construct a proper ADT + a `Buildable` instance.
2. `CofferError` is meant to be backend-agnostic, but it assumes that all backends will (possibly) throw a `ServantError`. However, this assumption is not true - the `pass` backend, for example, does not use servant.

The proposed solution is to refactor the `CofferError` such that it can wrap any backend-specific error:

```hs
data CofferError where
  ...
  BackendError :: Buildable err => err -> CofferError
```

And then create backend-specific error types, with proper constructors,and move the `ServantError` constructor:
```hs
data VaultError
  = ServantError ClientError
  | FieldMetadataNotFound EntryPath FieldName
  | ...
```

# Acceptance criteria

1. We're using ADTs + Buildable to model errors, and not just building ad-hoc strings with error messages.
2. `CofferError` is backend-agnostic, and the haddocks make this explicit.
3. We have backend-specific error types (at the moment, it's just `VaulError`)

DK318:
## Problem 
There are 2 problems here:
- In the `Backend.Vault.Kv` module, we're constructing `OtherError` with user-facing error messages as strings.
- `CofferError` is meant to be backend-agnostic, but it assumes that all backends will (possibly) throw a `ServantError`. However, the `pass` backend, for example, does not use servant.

## Solution
1. Used ADTs + Buildable to model errors.
2. Made `CofferError` backend-agnostic.
3. Created backend-specific error type class.

## Related issue(s)

- Fixes #72.

Fixed #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/lib/Backend/Commands.hs
+++ b/lib/Backend/Commands.hs
@@ -36,8 +36,7 @@ import Entry
   (Entry, EntryTag, Field, FieldKey, FieldVisibility(..), dateModified, fieldValue, fields,
   newEntry, newField, path, value, visibility)
 import Entry qualified as E
-import Error (CofferError(..))
-import Fmt (Builder, pretty, unlinesF)
+import Error (CofferError(..), InternalCommandsError(EntryPathDoesntHavePrefix, InvalidEntry))
 import GHC.Exts (Down(..), sortWith)
 import Polysemy
 import Polysemy.Error (Error, throw)
@@ -349,11 +348,9 @@ buildCopyOperations oldBackend newBackend oldQPath@(QualifiedPath oldBackendName
         newEntry <- entry & E.path . Path.entryPathParentDir %%~ \parentDir ->
           case Path.replacePathPrefix oldPath newPath parentDir of
             Just newParentDir -> pure newParentDir
-            Nothing -> throw $ OtherError $ unlinesF @_ @Builder
-              [ "Internal error:"
-              , "Expected path: '" <> pretty (entry ^. E.path) <> "'"
-              , "To have the prefix: '" <> pretty oldPath <> "'"
-              ]
+            Nothing -> do
+              let err = EntryPathDoesntHavePrefix (entry ^. E.path) oldPath
+              throw $ InternalCommandsError err
 
         let old = QualifiedPath oldBackendNameMb entry
         let new = QualifiedPath newBackendNameMb newEntry
@@ -530,12 +527,7 @@ getEntryOrDir backend path =
                   Nothing -> pure ()
 
               (_, Right subdir) -> go (rootPath <> subdir)
-              _ -> lift $ throw $ OtherError $ unlinesF
-                    [ "Internal error:"
-                    , "Backend returned a secret that is not a valid\
-                      \ entry or directory name."
-                    , "Got: '" <> secret <> "'."
-                    ]
+              _ -> lift $ throw $ InternalCommandsError (InvalidEntry secret)
 
 -- | This function gets all entries, that are exist in given entry path.
 --
--- a/lib/Backend/Vault/Kv.hs
+++ b/lib/Backend/Vault/Kv.hs
@@ -30,8 +30,8 @@ import Data.Text.Lazy qualified as TL
 import Data.Time (UTCTime)
 import Entry (Entry, Field, FieldKey, FieldValue(FieldValue), FieldVisibility)
 import Entry qualified as E
-import Error (CofferError(..))
-import Fmt
+import Error (BackendError, CofferError(..))
+import Fmt (Buildable(build), Builder, indentF, unlinesF, (+|), (|+))
 import GHC.Generics (Generic)
 import Network.HTTP.Client (defaultManagerSettings, newManager)
 import Network.HTTP.Client.TLS (tlsManagerSettings)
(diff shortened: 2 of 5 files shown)


MagicRB:
Id move `VaultError` into the backend folder, but aside from that LGTM

lib/CLI/Parser.hs
```diff
@@ -350,7 +350,7 @@ readEntryTag = do
   eitherReader \input ->
     newEntryTag (T.pack input) & first \err -> unlines
       [ "Invalid tag: " <> show input <> "."
-      , T.unpack err
+      , T.unpack (unBadEntryTag err)
```
dcastro, line 353: I think this should be the same as:

```suggestion
      , pretty err
```

lib/CLI/Parser.hs
```diff
@@ -376,7 +376,7 @@ readFieldKey' input = do
     Right tag -> pure tag
     Left err -> Left $ unlines
       [ "Invalid field name: " <> show input <> "."
-      , T.unpack err
+      , T.unpack (unBadFieldKey err)
```
dcastro, line 379: Same here

lib/Entry.hs
```diff
@@ -50,12 +52,17 @@ newtype FieldKey = UnsafeFieldKey Text
 keyCharSet :: [Char]
 keyCharSet = ['a'..'z'] ++ ['A'..'Z'] ++ ['0'..'9'] ++ "-_;"
 
-newFieldKey :: Text -> Either Text FieldKey
+newtype BadFieldKey = BadFieldKey { unBadFieldKey :: Text }
+
+instance Buildable BadFieldKey where
+  build (BadFieldKey err) = build err
```
dcastro, line 58:
```suggestion
newtype BadFieldKey = BadFieldKey { unBadFieldKey :: Text }
  deriving newtype Buildable
```

lib/Entry.hs
```diff
@@ -65,12 +72,17 @@ newtype EntryTag = UnsafeEntryTag Text
   deriving stock (Show, Eq, Ord)
   deriving newtype (A.ToJSON, A.FromJSON, Buildable)
 
-newEntryTag :: Text -> Either Text EntryTag
+newtype BadEntryTag = BadEntryTag { unBadEntryTag :: Text }
+
+instance Buildable BadEntryTag where
+  build (BadEntryTag err) = build err
```
dcastro, line 78:
```suggestion
newtype BadEntryTag = BadEntryTag { unBadEntryTag :: Text }
  deriving newtype BadEntryTag
```

dcastro (changes requested):

dcastro:
> Id move VaultError into the backend folder, but aside from that LGTM

Agreed. More specifically, the `Backend/Vault` folder.

Either to its own file or to the `Backend.Vault.Kv` module, whichever you prefer.

lib/CLI/Parser.hs
DK318: Done

lib/CLI/Parser.hs
DK318: Done

lib/Entry.hs

lib/Entry.hs

dcastro (approved):

lib/Backend/Commands.hs
```diff
@@ -350,11 +349,7 @@ buildCopyOperations oldBackend newBackend oldQPath@(QualifiedPath oldBackendName
         newEntry <- entry & E.path . Path.entryPathParentDir %%~ \parentDir ->
           case Path.replacePathPrefix oldPath newPath parentDir of
             Just newParentDir -> pure newParentDir
-            Nothing -> throw $ OtherError $ unlinesF @_ @Builder
-              [ "Internal error:"
-              , "Expected path: '" <> pretty (entry ^. E.path) <> "'"
-              , "To have the prefix: '" <> pretty oldPath <> "'"
-              ]
+            Nothing -> throw $ InternalCommandsError (EntryPathDoesntHavePrefix (entry ^. E.path) oldPath)
```
MagicRB, line 352: this line is a tad over 80, it could be broken up

lib/Backend/Vault/Kv.hs
```diff
@@ -54,6 +55,66 @@ data VaultKvBackend =
...
+        ]
+    ServantError (ConnectionError exception) ->
+      unlinesF @_ @Builder
+        [ "Connection error. No response was received."
+        , (build . show) exception
+        ]
+    FieldMetadataNotFound entryPath fieldName ->
+      "Could not find coffer metadata for field '" +| fieldName
+        |+ "' at '" +| entryPath |+ "'"
+    BadFieldNameError err -> build err
```
MagicRB, line 104: This is not real an error specific to Vault, we'll end up with duplication in the Pass backend.

lib/Backend/Vault/Kv.hs
```diff
@@ -54,6 +55,66 @@ data VaultKvBackend =
...
+    BadMasterFieldName key err ->
+      unlinesF @_ @Builder
+        [ "Attempted to create new field name from '" +| key |+ "'"
+        , ""
+        , build err
+        ]
+    BadEntryTagError err -> build err
```
MagicRB, line 111: Same thing as with `BadFieldNameError`

lib/Backend/Vault/Kv.hs
```diff
@@ -54,6 +55,66 @@ data VaultKvBackend =
...
+    ServantError (ConnectionError exception) ->
+      unlinesF @_ @Builder
+        [ "Connection error. No response was received."
+        , (build . show) exception
+        ]
+    FieldMetadataNotFound entryPath fieldName ->
+      "Could not find coffer metadata for field '" +| fieldName
+        |+ "' at '" +| entryPath |+ "'"
+    BadFieldNameError err -> build err
+    BadMasterFieldName key err ->
```
MagicRB, line 105: Same thing as with `BadFieldNameError`.

MagicRB (changes requested):

lib/Backend/Commands.hs

lib/Backend/Vault/Kv.hs

lib/Backend/Vault/Kv.hs

MagicRB (approved):

Issue #26, opened 2022-02-27
Setup project


dcastro:
We need the following:

- [ ] a buildkite pipeline
  - [ ] [reuse](https://reuse.software/) checks
  - [ ] trailing-whitespace-check
- [ ] code_style.md document
- [ ] makefiles
- [ ] support for cabal & stack
- [ ] tasty test suite
- [ ] github issue/pull request templates
- [ ] license files
- [ ] stylish-haskell
- [ ] hlint

PR #77, opened 2022-04-24
[#26] Fix `validate cabal files` CI step

dcastro:
## Description

Problem: The `validate cabal files` build step hasn't been working ever
since we added it. We can see this message in the logs:

```
./scripts/validate-cabal-files.sh: line 39: diff: command not found
```

The reason why the build step did not actually ever fail (thus, it was a false
negative) is because we temporarily use `set +e` to mask all non-zero
exit codes.

Solution: add the missing `diffTools` to the nix shell.

## Related issue(s)

Fixed part of #26 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


DK318 (approved):
LGTM

PR #78, opened 2022-04-28
[#22] Copy with equal paths behavior fix

### LINKED ISSUE #22: `coffer copy` with two equal paths behavior
In #13 next behavior was proposed: throwing an error (like `cp` command), regardless of whether `-f` was used.

sancho20021:
### Problem
Currently copying entry / directory to itself is allowed in force mode.

This copying is very unnecessary, we can't see such behavior in
standard Unix `cp` command. The same applies to renaming
entries to it's current name.

### Solution
Check qualified paths for equality before copying.

## Related issue(s)

Fixed #22 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -130,6 +130,7 @@ main = do
           CPRCreateErrors errors -> do
             errorMsgs <- buildErrorMessages errors
             printError $ unlinesF @_ @Builder $ "The following entries cannot be renamed:" : "" : errorMsgs
+          CPRSamePath path -> samePaths path
 
       SomeCommand cmd@(CmdCopy opts) -> do
         runCommand config cmd >>= \case
@@ -144,6 +145,7 @@ main = do
           CPRCreateErrors errors -> do
             errorMsgs <- buildErrorMessages errors
             printError $ unlinesF @_ @Builder $ "The following entries cannot be copied:" : "" : errorMsgs
+          CPRSamePath path -> samePaths path
 
       SomeCommand cmd@(CmdDelete opts) -> do
         runCommand config cmd >>= \case
@@ -187,6 +189,10 @@ main = do
 
       pathNotFound :: Member (Embed IO) r => QualifiedPath Path -> Sem r ()
       pathNotFound path = printError $ "Entry or directory not found at '" +| path |+ "'."
+
+      samePaths :: Member (Embed IO) r => QualifiedPath Path -> Sem r ()
+      samePaths path =
+        printError $ "'" +| path |+ "' and '" +| path |+ "' are the same path."
 
       createErrorToBuilder :: CreateError -> Builder
       createErrorToBuilder = \case
--- a/lib/Backend/Commands.hs
+++ b/lib/Backend/Commands.hs
@@ -314,8 +314,17 @@ buildCopyOperations
   :: forall r
    . (Members '[BackendEffect, Embed IO, Error CofferError, Error CopyResult] r)
   => SomeBackend -> SomeBackend -> QualifiedPath Path -> QualifiedPath Path -> Bool -> Sem r [CopyOperation]
-buildCopyOperations oldBackend newBackend oldQPath@(QualifiedPath oldBackendNameMb oldPath) (QualifiedPath newBackendNameMb newPath) force = do
+buildCopyOperations
+  oldBackend
+  newBackend
+  oldQPath@(QualifiedPath oldBackendNameMb oldPath)
+  newQPath@(QualifiedPath newBackendNameMb newPath)
+  force
+    = do
   entryOrDir <- getEntryOrDirThrow oldBackend CPRPathNotFound oldQPath
+
+  -- Don't need to copy directory or entry to itself
+  when (oldQPath == newQPath) $ throw (CPRSamePath oldQPath)
 
   -- Build a list of operations to perform.
   nowUtc <- embed getCurrentTime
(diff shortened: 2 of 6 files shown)


lib/Backend/Commands.hs
```diff
@@ -319,6 +319,10 @@ buildCopyOperations
 buildCopyOperations oldBackend newBackend oldQPath@(QualifiedPath oldBackendNameMb oldPath) (QualifiedPath newBackendNameMb newPath) force = do
   entryOrDir <- getEntryOrDirThrow oldBackend CPRPathNotFound oldQPath
 
+  -- Don't need to copy directory or entry to itself
+  when ((oldBackendNameMb == newBackendNameMb) && (oldPath == newPath)) $
```
DK318, line 323: We could simplify this check like this
````haskell
when (oldQPath == newQPath) do
...
````

File: tests/golden/copy-command/copy-command.bats [hunk 2]
```diff
@@ -277,3 +277,41 @@ EOF
...
+
+@test "copy an entry to itself" {
+  coffer create /a
+
+  run coffer copy -f /a /a
+
+  assert_failure
+  assert_output "[ERROR] '/a' and '/a' are the same path."
+
+  run cleanOutput
```
DK318, line 289: We don't need `cleanOutput` here. It just replaces all timestamps in the output

DK318 (changes requested):
Please turn on trimming trailing whitespace in your editor.

lib/Backend/Commands.hs
sancho20021: Done

tests/golden/copy-command/copy-command.bats
sancho20021: Ok, removed

sancho20021:
> 

Yes, thanks! Didn't have that in my configuration

dcastro:
There's an issue with the xrefcheck step in the pipeline, @DK318 is working on it (see #79).

Once his PR is merged, please rebase your branch on top of main, and the pipeline for your PR should start passing.

tests/golden/copy-command/copy-command.bats
```diff
@@ -313,3 +313,26 @@ EOF
     b - [2000-01-01 01:01:01]
 EOF
 }
+
+@test "equal paths copy from one backend to the other" {
+  coffer create /a/b
+
+  run coffer copy /a second#/a
```
DK318, line 320: Hmm. Maybe we need the test like this
````
coffer copy vault-local#/a vault-local#/a -f
````

**Review by DK318 — COMMENTED** [review]
LGTM

File: tests/golden/copy-command/copy-command.bats (hunk 3 above)
Comment (sancho20021) [reply tests/golden/copy-command/copy-command.bats:335]: Done. Btw, this test showed that I forgot to include backend names when showing paths in the error message. Now it is fixed

File: lib/CLI/Types.hs [hunk 4]
```diff
@@ -73,7 +73,7 @@ data CopyResult
   = CPRSuccess [(QualifiedPath EntryPath, QualifiedPath EntryPath)]
   | CPRPathNotFound (QualifiedPath Path)
   | CPRMissingEntryName
-  | CPRSamePath Path Path
+  | CPRSamePath (QualifiedPath Path) Path
```
DK318, line 76: Why not `CPRSamePath (QualifiedPath Path) (QualifiedPath Path)`?

app/Main.hs
```diff
@@ -190,8 +190,10 @@ main = do
       pathNotFound :: Member (Embed IO) r => QualifiedPath Path -> Sem r ()
       pathNotFound path = printError $ "Entry or directory not found at '" +| path |+ "'."
 
-      samePaths :: Member (Embed IO) r => Path -> Path -> Sem r ()
-      samePaths old new = printError $ "'" +| old |+ "' and '" +| new |+ "' are the same path."
+      samePaths :: Member (Embed IO) r => QualifiedPath Path -> Path -> Sem r ()
+      samePaths old@(QualifiedPath backend _) newPath =
+        printError $ "'" +| old |+ "' and '" +|
+          (QualifiedPath backend newPath) |+ "' are the same path."
```
DK318, line 196:
```suggestion
      samePaths :: Member (Embed IO) r => QualifiedPath Path -> QualifiedPath Path -> Sem r ()
      samePaths oldPath newPath =
        printError $ "'" +| oldPath |+ "' and '" +|
          newPath |+ "' are the same path."
```

app/Main.hs
```diff
@@ -188,6 +190,11 @@ main = do
       pathNotFound :: Member (Embed IO) r => QualifiedPath Path -> Sem r ()
       pathNotFound path = printError $ "Entry or directory not found at '" +| path |+ "'."
 
+      samePaths :: Member (Embed IO) r => QualifiedPath Path -> Path -> Sem r ()
+      samePaths old@(QualifiedPath backend _) newPath =
+        printError $ "'" +| old |+ "' and '" +|
+          (QualifiedPath backend newPath) |+ "' are the same path."
```
dcastro, line 196: Instead of extracting the backend from `oldQPath :: QualifiedPath Path` and adding it to `newPath :: Path`, we could just save the qualified `newQPath :: QualifiedPath Path` inside `CPRSamePath`?

Or even just save one of the two paths (since they're the same anyway)?

sancho20021: Oh, really, I will just leave one of them. I think in the beginning I decided to pass two, because I thought they might have different string representations 🤣

lib/CLI/Types.hs
sancho20021: Removed second argument

app/Main.hs
sancho20021: Removed second argument

Issue #44, opened 2022-03-21
Merge `--filter` and `--filter-field` into one option in core, CLI and web API


Heimdell:
# Clarification and motivation

@dcastro: I think the --filter and --filter-field options can be merged into a single --filter option.

e.g.:
```
--filter name~vault               <- filter entries by entry name
--filter url:value~google.com     <- filter entries by field value
```

Probably, multiple `--filter` in CLI should act all at once, and so do `filter[]=...` in web API.

# Acceptance criteria

- [ ] core has them merged
- [ ] CLI has them merged
- [ ] web API has them merger

dcastro:
Looks like the core + cli have been updated in #94, and the web API has been updated during the last rebase in #37. Closing this issue now.

PR #94, opened 2022-04-29
[#44] Merge `--filter` and `--filter-field` options

sancho20021:
### Problem

@Heimdell in #44 suggested to merge --filter and --filter-field options into a single --filter option.

e.g.:
```
--filter name~vault               <- filter entries by entry name
--filter url:value~google.com     <- filter entries by field value
```

### Solution

Modified `parseFilter :: MParser Filter` supplementing it with `parseFilterByField`'s functionality.

Changed related datatypes so there is no separation between `filter` and `filter-field`.

## Related issue(s)

- Fixed part of #44:
  - [x] core has filters merged
  - [x] CLI has filters merged 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/lib/CLI/Parser.hs
+++ b/lib/CLI/Parser.hs
@@ -35,6 +35,7 @@ import Entry
 import Fmt (pretty)
 import Options.Applicative
 import Options.Applicative.Help.Pretty qualified as Pretty
+import Text.Megaparsec (try)
 import Text.Megaparsec qualified as P
 import Text.Megaparsec.Char qualified as P
 import Text.Megaparsec.Char.Lexer qualified as P
@@ -231,14 +232,6 @@ findOptions =
           , helpDoc $ Just $ Pretty.vsep
             [ "Filter entries."
             , expectedFilterFormat
-            ]
-          ])
-    <*> many (option readFilterField $ mconcat
-          [ metavar "FILTERFIELD"
-          , long "filter-field"
-          , helpDoc $ Just $ Pretty.vsep
-            [ "Filter entries based on a field."
-            , expectedFilterFieldFormat
             ]
           ])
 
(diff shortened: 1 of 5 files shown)


lib/CLI/Parser.hs
```diff
@@ -505,29 +498,11 @@ expectedQualifiedPathFormat = Pretty.vsep
...
-      , ""
-      , "Parser error:"
-      , P.errorBundlePretty err
-      ]
-
-expectedFilterFieldFormat :: Pretty.Doc
-expectedFilterFieldFormat = Pretty.vsep
-  [ "Expected format is: '<fieldname>:contents~<substring>' or `<fieldname>:date<op><date>`."
+  [ "Expected format is: 'name~<substring>' or 'date<op><date>',"
+  , "'<fieldname>:contents~<substring>' or '<fieldname>:date<op><date>'."
```
dcastro, line 502: Hmm, now that we support 4 different formats, maybe displaying them in a bulleted list would look better? And a short description of what which format does.

See the example in `expectedSortFormat`.

dcastro (commented):
LGTM :+1: Just one small comment

lib/CLI/Parser.hs
sancho20021: Done. But I noticed that the error messages when parsing `filter` option differ from `sort`.
When using `--sort`, we cannot see any parsing exceptions in raw form, e.g.:
```
>>> coffer find --sort name:
option --sort: Invalid direction: ''.
Choose one of: 'asc', 'desc'.


Usage: coffer find [PATH] [TEXT] [--sort SORT] [--filter FILTER]
  Find and list entries, optionally filtering
```

But in case of `--filter` the raw parse exceptions appear (and it was before this pull request), e.g.:
```
>>> coffer find --filter FILTER
...
Parser error:
1:7:
  |
1 | FILTER
  |       ^
unexpected end of input
expecting ':' or fieldname
...
```
I think that it would be nice if we chose one style of reporting errors.

lib/CLI/Parser.hs
```diff
@@ -588,7 +563,7 @@ parseFilterOp =
 
 parseFilter :: MParser Filter
 parseFilter =
-  parseFilterByName <|> parseFilterByDate
+  try parseFilterByName <|> try parseFilterByDate <|> parseFilterByField
```
DK318, line 569: No need for `try` here. All these formats start with different strings.

DK318 (approved):
Just one small comment. LGTM

lib/CLI/Parser.hs
sancho20021: Tests do not pass without try. `filter field with name 'date' and 'name'` in particular. This happens when field names are equal to 'date' or 'name'.

DK318: Makes sense :thinking:

lib/CLI/Parser.hs
dcastro: Ah, that seems to be because we're using megaparsec to parse filters (which have a more complex syntax) and `Text.splitOn ":"` to parse sorts (which have a simpler syntax).

Yeah, we could use megaparsec for both, for consistency. Please create a separate issue for that.

dcastro (approved):

PR #97, opened 2022-05-02
Replace current CI implementation with Nix-flake based CI

notgne2:
## Description

We have newer practices for CI systems, so we should probably utilize them here, including our newest addition: `mkPipeline`, which generates individual pipeline steps for CI out of your flake, as opposed to lumping them all into a `nix flake check` step.

## Related issue(s)

#26 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


notgne2:
(might take some messing around with, this is my first time setting something up with mkPipeline)

notgne2:
that seems to all be working except for weeder, which fails with
>   weeder: user error (Didn't find exactly 1 cabal file in .)

I might leave figuring that out to somebody that knows more about cabal

notgne2:
Setting as draft until https://github.com/serokell/serokell.nix/pull/69 is addressed, then will need to clean up commits

dcastro:
I noticed that we're missing a few steps we used to have (there may be more):
* `golden tests`
* `validate cabal files`
* `stylish`
* `lint` (there is a new `hlilnt` step now, but looking at the logs it doesn't seem like it's actually running `hlint`)
* `shellcheck`
* `linux static executable`
* `xrefcheck`

Also:
* the `tests` step seems to be building the tests component, but not actually running it
* same for the `doctests` step (In fact, I had issues trying to get `doctests` to run correctly on the CI. so if we can't get it to run and pass, we might as well just ignore it in the CI, which is what we've been doing so far).


Questions: 
* How do I run these CI steps locally? Previously, I was able to copy/paste the commands from `.buildkite/pipeline.yml`. But I'm not familiar with flakes, so I wouldn't know how to reproduce this locally.

notgne2:
I'll add in the missing test-runnig and try to add all the missing tests too

> How do I run these CI steps locally?

With using mkPipeline, it's generating our pipeline steps from what is under `packages` and `checks`, the same things that will be built by Nix if you run `nix flake check`.

notgne2:
Adding xrefcheck would be ugly unless we can first deal with https://github.com/serokell/xrefcheck/issues/54 (which would probably ending up looking exactly like this PR)

notgne2:
It also seems like "validate cabal files" might be tedious to add, I tried it and it failed while attempting network access, is this because I configured it wrong? or is that step just impure

dcastro:
> It also seems like "validate cabal files" might be tedious to add, I tried it and it failed while attempting network access, is this because I configured it wrong? or is that step just impure

Looks like it's impure. The script `validate-cabal-files.sh` calls [stack2cabal](https://github.com/hasufell/stack2cabal), which I think needs to fetch the stack resolver.

PR #99, opened 2022-05-06
[#83] Web API proper status codes

### LINKED ISSUE #83: Return proper status codes when an error occurs

When a command fails, the Web API returns a 200 status code with an error message in the body. For example, creating the same entry twice results in:

```hs
$ curl -s -D /dev/stderr \
  -H 'token: root' \
  'localhost:8081/api/v1/content/create?path=/ex1' \
  -X POST \
  -H 'Content-Type: application/json' \
  -d '[{"a": "b"}, {"c": "d"}]' | jq

HTTP/1.1 200 OK
Transfer-Encoding: chunked
Date: Fri, 29 Apr 2022 10:38:15 GMT
Server: Warp/3.3.18
Content-Type: application/json;charset=utf-8

{
  "contents": {
    "unEntryPath": [
      "ex1"
    ]
  },
  "tag": "CREntryAlreadyExists"
}

```

We should return an appropriate status code depending on the error (in this specific case, it should be a 409 Conflict) with an error message in the body:

```json
{
  "error": "An entry already exists at <path>",
  "code": 123
}
```

--- 

Each type of error should also be associated with a numeric code.
This will allow the frontend to know exactly what went wrong and act appropriately.
For example, when a copy operation fails, it may return something like:

```
409 Conflict
{
  "error": "An entry already exists at <path>",
  "code": 123
}
```

Or:

```
409 Conflict
{
  "error": "<path> is a directory",
  "code": 456
}
```

And the frontend will be able to match on the numeric code and decide what to do next.

We should create a markdown table in `/docs` with all the error codes and what they represent.

# Acceptance criteria

* Endpoints return an appropriate status code when they fail
* The response body contains an error message and a numeric code
* We have markdown table explaining what the numeric error codes represent

DK318:
## Description

Web API proper status codes + `coffer` commands error codes.

## Related issue(s)

- Fixes #83 

Fixed #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/app/cli/Main.hs
+++ b/app/cli/Main.hs
@@ -10,15 +10,14 @@ import Backend.Interpreter
 import CLI.Parser
 import CLI.PrettyPrint
 import CLI.Types
-import Coffer.Directory qualified as Dir
-import Coffer.Path (EntryPath, Path, QualifiedPath(qpPath))
+import Coffer.PrettyPrint
+  (PrettyPrintMode(CLI), buildCopyResult, buildCreateResult, buildDeleteFieldResult,
+  buildDeleteResult, buildRenameResult, buildSetFieldResult, buildTagResult, buildViewResult)
 import Config (Config(..), configCodec)
 import Control.Lens
-import Control.Monad (forM_, when)
 import Data.Maybe (fromMaybe)
 import Data.Text (pack)
 import Data.Text.IO qualified as TIO
-import Entry qualified as E
 import Error
 import Fmt
 import Options.Applicative (execParser)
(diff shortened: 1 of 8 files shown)


lib/CLI/Types.hs
```diff
@@ -376,3 +471,47 @@ parseFilterDate = do
...
+      then [[int|s|
+        These actions would be done:
+        #{messages}
+      |]]
+      else [messages]
+  CPRPathNotFound path -> [buildPathNotFound path]
+  CPRMissingEntryName ->
+    ["The destination path is not a valid entry path. Please specify the new name of the entry."]
+  CPRCreateErrors errors -> buildErrorMessages errors
+  CPRSamePath path -> [buildSamePath path]
```
sancho20021, line 517: Maybe create two separate functions not to pass bool every time?
```

buildCopyResult :: CopyResult -> [Builder]
buildCopyResult = buildCopyOrRenameResult False

buildRenameResult :: RenameResult -> [Builder]
buildRenameResult = buildCopyOrRenameResult True
```

lib/Web/Server.hs
```diff
@@ -42,6 +47,89 @@ url = BaseUrl Http "127.0.0.1" 8200 ""
...
+  res@SFREntryNotFound{} -> throwError err404 { errBody = cofferServerErrorToBytes (pretty res) 300 }
+  res@SFRMissingFieldContents{} -> throwError err400 { errBody = cofferServerErrorToBytes (pretty res) 301 }
+
+throwDeleteFieldError :: DeleteFieldResult -> Handler a
+throwDeleteFieldError = \case
+  DFRSuccess{} -> throwError err500 { errBody = "Expected delete-field error, got DFRSuccess." }
+  res@DFREntryNotFound{} -> throwError err404 { errBody = cofferServerErrorToBytes (pretty res) 400 }
+  res@DFRFieldNotFound{} -> throwError err404 { errBody = cofferServerErrorToBytes (pretty res) 401 }
+
+throwCopyOrRenameError :: Bool -> CopyResult -> Handler a
```
sancho20021, line 93: Same as with `buildCopyOrRenameResult`

app/cli/Main.hs
```diff
@@ -71,128 +67,83 @@ main = do
...
-          DRSuccess paths -> do
-            when (doDryRun opts) do
-              pprint "These actions would be done:"
-            forM_ paths \path ->
-              printSuccess [int|s|Deleted '#{path}'.|]
-
-      SomeCommand cmd@(CmdTag opts) -> do
+          res@DRPathNotFound{} -> printError $ build res
+          res@DRDirectoryFound{} -> printError $ build res
+          res@DRSuccess{} -> pprint $ build res
```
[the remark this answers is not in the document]
DK318: Yeah, it's correct. See `Buildable` instance in `Types` module

lib/Web/Server.hs
```diff
@@ -59,7 +147,7 @@ reportErrors io = do
       throwError err500 { errBody = errorToServantErrorMsg ioError }
```
dcastro, line 147: By the way, in `errorToServantErrorMsg`, we should use `displayException` instead of `show`. It usually displays a more human-readable version of the exception.

dcastro: Also, instead of going
```
String -> Strict.Text -> Strict.ByteString -> Lazy.ByteString
```

We can go
```
String -> Lazy.Text -> Lazy.ByteString
```

Using `Data.Text.Lazy.pack` and `Data.Text.Lazy.Encoding.encodeUtf8`

lib/Web/Server.hs
```diff
@@ -42,6 +47,89 @@ url = BaseUrl Http "127.0.0.1" 8200 ""
 mount :: Text
```
dcastro, line 47: Unrelated to the PR, but while we're here: can you please delete this? It's not being used anywhere.

lib/Web/Server.hs
```diff
@@ -42,6 +47,89 @@ url = BaseUrl Http "127.0.0.1" 8200 ""
...
+    createErrorToCode = \case
+      CEParentDirectoryIsEntry{} -> 503
+      CEDestinationIsDirectory{} -> 504
+      CEEntryAlreadyExists{} -> 505
+
+throwDeleteError :: DeleteResult -> Handler a
+throwDeleteError = \case
+  DRSuccess{} -> throwError err500 { errBody = "Expected delete error, got DRSuccess." }
+  res@DRPathNotFound{} -> throwError err404 { errBody = cofferServerErrorToBytes (pretty res) 600 }
+  res@DRDirectoryFound{} -> throwError err404 { errBody = cofferServerErrorToBytes (pretty res) 601 }
```
dcastro, line 124: This one's not a 404. But I'm not sure what the appropriate status code should be here either...
Hmm let's just default to 400.

dcastro (changes requested):

lib/Web/Server.hs
DK318: I think I should do the same with `copyOrRenameHandler`, not with this one

lib/Web/Server.hs
```diff
@@ -42,6 +47,89 @@ url = BaseUrl Http "127.0.0.1" 8200 ""
...
+  CRSuccess{} -> throwError err500 { errBody = "Expected create error, got CRSuccess." }
+  CRCreateError createError ->
+    case createError of
+      res@CEParentDirectoryIsEntry{} -> throwError err409 { errBody = cofferServerErrorToBytes (pretty res) 200 }
+      res@CEDestinationIsDirectory{} -> throwError err409 { errBody = cofferServerErrorToBytes (pretty res) 201 }
+      res@CEEntryAlreadyExists{} -> throwError err409 { errBody = cofferServerErrorToBytes (pretty res) 202 }
+
+throwSetFieldError :: SetFieldResult -> Handler a
+throwSetFieldError = \case
+  SFRSuccess{} -> throwError err500 { errBody = "Expected set-field error, got SFRSuccess." }
```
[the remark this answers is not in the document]
DK318: Don't get what you mean

lib/Web/Server.hs
```diff
@@ -42,6 +47,89 @@ url = BaseUrl Http "127.0.0.1" 8200 ""
 mount :: Text
 mount = "secret"
 
+data CofferServerError = CofferServerError
+  { cseError :: Text
+  , cseCode :: Int
+  }
+  deriving stock (Show, Generic)
+deriveJSON (aesonPrefix pascalCase) ''CofferServerError
```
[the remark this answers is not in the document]
DK318: Done

lib/Web/Server.hs
```diff
@@ -42,6 +47,89 @@ url = BaseUrl Http "127.0.0.1" 8200 ""
...
+  res@DFREntryNotFound{} -> throwError err404 { errBody = cofferServerErrorToBytes (pretty res) 400 }
+  res@DFRFieldNotFound{} -> throwError err404 { errBody = cofferServerErrorToBytes (pretty res) 401 }
+
+throwCopyOrRenameError :: Bool -> CopyResult -> Handler a
+throwCopyOrRenameError rename = \case
+  CPRSuccess{} -> throwError err500 { errBody = "Expected " <> cmdName <> " error, got CPRSuccess." }
+  res@CPRPathNotFound{} -> throwError err404 { errBody = cofferServerErrorToBytes (prettySingleMessage res) 500 }
+  res@CPRMissingEntryName{} -> throwError err400 { errBody = cofferServerErrorToBytes (prettySingleMessage res) 501 }
+  res@CPRSamePath{} -> throwError err409 { errBody = cofferServerErrorToBytes (prettySingleMessage res) 502 }
+  res@(CPRCreateErrors createErrors) -> do
```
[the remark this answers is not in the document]
DK318: Done

lib/Web/Server.hs
```diff
@@ -29,9 +29,14 @@ import Web.API
 import Backend.Interpreter (runBackend)
```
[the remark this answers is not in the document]
DK318: Done

lib/Web/Server.hs
```diff
@@ -42,6 +47,89 @@ url = BaseUrl Http "127.0.0.1" 8200 ""
...
+  CPRSuccess{} -> throwError err500 { errBody = "Expected " <> cmdName <> " error, got CPRSuccess." }
+  res@CPRPathNotFound{} -> throwError err404 { errBody = cofferServerErrorToBytes (prettySingleMessage res) 500 }
+  res@CPRMissingEntryName{} -> throwError err400 { errBody = cofferServerErrorToBytes (prettySingleMessage res) 501 }
+  res@CPRSamePath{} -> throwError err409 { errBody = cofferServerErrorToBytes (prettySingleMessage res) 502 }
+  res@(CPRCreateErrors createErrors) -> do
+    let messages = buildCopyOrRenameResult rename res <&> pretty
+    let codes = createErrors <&> createErrorToCode . snd
+    let serverErrors = zipWith CofferServerError messages codes
+    let body = encodingToLazyByteString $ toEncoding serverErrors
+    throwError err409 { errBody = body }
```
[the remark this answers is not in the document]
DK318: Done

lib/Web/Server.hs
DK318: Done

lib/Web/Server.hs
DK318: Done

lib/Web/Server.hs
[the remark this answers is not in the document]
dcastro: I mean, the only reason why this error 500 might happen is because we've separated the code like this:

```hs
public run token sfoQPath sfoFieldName  =
  run token ... >>= \case
      res@SFRSuccess{} -> pure res
      res@SFREntryNotFound{} -> throwSetFieldError res
      res@SFRMissingFieldContents{} -> throwSetFieldError res
```
```hs
throwSetFieldError :: SetFieldResult -> Handler a
throwSetFieldError = \case
  SFRSuccess{} -> throwError err500 ...
  res@SFREntryNotFound{} -> throwError $ addApplicationJSONHeader err404 ...
  res@SFRMissingFieldContents{} -> throwError $ addApplicationJSONHeader err400 ...
```

--------

There are two ways of eliminating this branch. One is by inlining `throwSetFieldError`:

```hs
public run token sfoQPath sfoFieldName  =
  run token ... >>= \case
      res@SFRSuccess{} -> pure res
      res@SFREntryNotFound{} -> throwError $ addApplicationJSONHeader err404
      res@SFRMissingFieldContents{} -> throwError $ addApplicationJSONHeader err404
```

Another way is by _not_ pattern matching on the `*Result` before calling `throwSetFieldError` (and also renaming it to simply `handleSetFieldResult`):

```hs
public run token sfoQPath sfoFieldName  =
  run token ... >>= handleSetFieldResult
```
```hs
handleSetFieldResult :: handleSetFieldResult -> Handler a
handleSetFieldResult = \case
  res@SFRSuccess{} -> pure res
  res@SFREntryNotFound{} -> throwError $ addApplicationJSONHeader err404 ...
  res@SFRMissingFieldContents{} -> throwError $ addApplicationJSONHeader err400 ...
```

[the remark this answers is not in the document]
dcastro: In the case of `throwSetFieldError`, it's used in the implementation of 3 endpoints, so option (2) is better.

But for most cases (like `throwDeleteFieldError` and `throwViewError`), they're only used in 1 endpoint each, so inlining them is better.

lib/Web/Server.hs
```diff
@@ -29,18 +28,142 @@ import Web.API
...
+
+throwTagError :: TagResult -> Handler a
+throwTagError = \case
+  TRSuccess{} -> throwError err500 { errBody = "Expected tag error, got TRSuccess." }
+  res@TREntryNotFound{} ->
+    throwError $ addApplicationJSONHeader err404 { errBody = cofferServerErrorToBytes (pretty res) 700 }
+  res@TRTagNotFound{} ->
+    throwError $ addApplicationJSONHeader err404 { errBody = cofferServerErrorToBytes (pretty res) 701 }
+  res@TRDuplicateTag{} ->
+    throwError $ addApplicationJSONHeader err409 { errBody = cofferServerErrorToBytes (pretty res) 702 }
```
dcastro, line 164: We should encapsulate this pattern in a single function and use it everywhere, e.g.:

```hs
throwCofferServerError :: ServerError -> Int -> Text -> Handler a
```

This single function would be responsible for creating the json body and adding the content-type header.

dcastro: And a more generic version of it, to use in `CPRCreateErrors`
```hs
throwCofferServerErrors :: ServerError -> [(Int, Text)] -> Handler a
```

lib/CLI/PrettyPrint.hs
```diff
@@ -98,3 +99,12 @@ buildField (fk, field) = do
...
   "[" <> build (formatTime defaultTimeLocale "%F %T" time) <> "]"
+
+buildPathNotFound :: QualifiedPath Path -> Builder
+buildPathNotFound path = [int|s|Entry or directory not found at '#{path}'.|]
+
+buildEntryNotFound :: QualifiedPath EntryPath -> Builder
+buildEntryNotFound entryPath = [int|s|Entry not found at '#{entryPath}'.|]
+
+buildSamePath :: QualifiedPath Path -> Builder
+buildSamePath path = [int|s|'#{path}' and '#{path}' are the same path.|]
```
dcastro, line 110: These 3 functions can also be moved to `Coffer.PrettyPrint`

lib/Web/Server.hs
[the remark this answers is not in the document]
DK318: Done

lib/Web/Server.hs
DK318: Done

lib/CLI/PrettyPrint.hs
DK318: Done

lib/Web/Server.hs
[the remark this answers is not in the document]
dcastro: > But for most cases (like `throwDeleteFieldError` and `throwViewError`), they're only used in 1 endpoint each, so inlining them is better.

I think you forgot this bit.

dcastro (changes requested):

lib/Web/Server.hs
```diff
@@ -28,19 +27,149 @@ import Web.API
...
+
+addApplicationJSONHeader :: ServerError -> ServerError
+addApplicationJSONHeader err@(ServerError _ _ _ headers) =
+  err { errHeaders = applicationJSONHeader : headers }
+  where
+    applicationJSONHeader = ("Content-Type", "application/json;charset=utf-8")
+
+throwCofferServerError :: ServerError -> Int -> Text -> Handler a
+throwCofferServerError err code msg =
+  throwError $ addApplicationJSONHeader err { errBody = cofferServerErrorToBytes msg code }
```
dcastro, line 65:
```suggestion
  throwCofferServerErrors err [(code, msg)]
```

lib/Web/Server.hs
```diff
@@ -28,19 +27,149 @@ import Web.API
...
+
+throwCofferServerErrors :: ServerError -> [(Int, Text)] -> Handler a
+throwCofferServerErrors err codesAndMsgs = do
+  let cofferServerErrors = uncurry (flip CofferServerError) <$> codesAndMsgs
+  throwError $ addApplicationJSONHeader err { errBody = encode cofferServerErrors }
```
dcastro, line 70: `addApplicationJSONHeader` is only used in one place, so i think it would make sense to simplify things and inline it here:

```suggestion
  throwError err
    { errBody = encode cofferServerErrors
    , errHeaders = ("Content-Type", "application/json;charset=utf-8") : errHeaders err
    }
```

lib/Web/Server.hs
dcastro: Also, we can delete `cofferServerErrorToBytes` after this change.

dcastro (changes requested):

lib/Web/Server.hs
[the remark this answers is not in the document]
DK318: Done

app/cli/Main.hs
```diff
@@ -71,128 +70,83 @@ main = do
...
-            [int|s|Entry already has the tag '#{tag}'.|]
+          res@(TREntryNotFound _) -> printError $ buildTagResult CLI res
+          res@TRSuccess{} -> printSuccess $ buildTagResult CLI res
+          res@TRTagNotFound{} -> printError $ buildTagResult CLI res
+          res@TRDuplicateTag{} -> printError $ buildTagResult CLI res
     where
       -- | Pretty-print a message.
       pprint :: Member (Embed IO) r => Builder -> Sem r ()
-      pprint = embed . putStrLn . fmt
+      pprint = embed . putStr . fmt
```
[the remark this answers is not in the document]
DK318: Done

docs/web-api-error-codes.md
```diff
@@ -0,0 +1,67 @@
+<!--
+ - SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io>
+ -
+ - SPDX-License-Identifier: MPL-2.0
+ -->
+
+# WEB API Error Codes
+
+These codes are grouped in the next way:
```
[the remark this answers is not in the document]
DK318: Done

docs/web-api-error-codes.md
```diff
@@ -0,0 +1,67 @@
...
+- `view` errors: 100-102
+- `create` errors: 200-202
+- `set-field` errors: 300-301
+- `delete-field` errors: 400-401
+- `rename`/`copy` errors: 500-505
+- `delete` errors: 600-601
+- `tag` errors: 700-702
+
+See next tables for detailed explanations of error codes.
```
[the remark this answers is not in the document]
DK318: Done

lib/Web/Server.hs
```diff
@@ -59,17 +188,16 @@ reportErrors io = do
       throwError err500 { errBody = errorToServantErrorMsg ioError }
 
     Right (Left e) -> do
-      throwError err500 { errBody = errorToServantErrorMsg (pretty @_ @Text e) }
+      throwError err500 { errBody = pretty e }
```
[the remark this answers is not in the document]
DK318: Done

lib/Web/Server.hs
DK318: Done

lib/Web/Server.hs
DK318: Done

lib/Web/Server.hs
dcastro: Forgot about this bit:

> Also, we can delete `cofferServerErrorToBytes` after this change.

DK318: Yeah. Removed

dcastro (approved):

PR #100, opened 2022-05-10
[#20] Rename effects in BackendEffect from `Secret` to `Entry`

### LINKED ISSUE #20: Rename effects in the BackendEffect from `Secret` to`Entry`
`readSecret` and so on don't precisely reflect the terminology we have agreed on. Changing them now would be a rather large change which would effect the whole codebase. Therefore we should wait for the `"big merge" until we tackle this.

sancho20021:
## Description

Use more suitable names for effects (~secret~ entry).

## Related issue(s)

Fixed #20 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/lib/Backend.hs
+++ b/lib/Backend.hs
@@ -4,10 +4,10 @@
 
 module Backend
   ( BackendEffect (..)
-  , readSecret
-  , writeSecret
-  , listSecrets
-  , deleteSecret
+  , readEntry
+  , writeEntry
+  , listDirectoryContents
+  , deleteEntry
   , Backend (..)
   , SomeBackend (..)
   , Effects
@@ -28,10 +28,10 @@ type Effects r = (Member (Embed IO) r, Member (Error CofferError) r)
 class Show a => Backend a where
   _name :: a -> BackendName
   _codec :: Toml.TomlCodec a
-  _writeSecret :: Effects r => a -> Entry -> Sem r ()
-  _readSecret :: Effects r => a -> EntryPath -> Sem r (Maybe Entry)
-  _listSecrets :: Effects r => a -> Path -> Sem r (Maybe [Text])
-  _deleteSecret :: Effects r => a -> EntryPath -> Sem r ()
+  _writeEntry :: Effects r => a -> Entry -> Sem r ()
+  _readEntry :: Effects r => a -> EntryPath -> Sem r (Maybe Entry)
+  _listDirectoryContents :: Effects r => a -> Path -> Sem r (Maybe [Text])
+  _deleteEntry :: Effects r => a -> EntryPath -> Sem r ()
 
 data SomeBackend where
   SomeBackend :: Backend a => a -> SomeBackend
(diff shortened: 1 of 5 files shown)


lib/Backend/Vault/Kv.hs
```diff
@@ -303,7 +303,7 @@ kvDeleteSecret backend path = do
...
   _name kvBackend = vbName kvBackend
   _codec = vaultKvCodec
-  _writeSecret = kvWriteSecret
-  _readSecret = kvReadSecret
-  _listSecrets = kvListSecrets
-  _deleteSecret = kvDeleteSecret
+  _writeEntry = kvWriteSecret
+  _readEntry = kvReadSecret
+  _listEntries = kvListSecrets
+  _deleteEntry = kvDeleteSecret
```
DK318, line 309: Please, rename these functions (`kvWriteSecret` and etc)

DK318 (changes requested):

lib/Backend/Vault/Kv.hs
sancho20021: Done. There are corresponding functions in `Vault.Internal`, but I'm not sure if I need to change those.

DK318: I think we want to unify these namings. So we should perform `secret` -> `entry` renaming

sancho20021: Done

DK318: Makes sense

lib/Error.hs
```diff
@@ -37,7 +37,7 @@ instance Buildable InternalCommandsError where
   build = \case
     InvalidEntry entry ->
       [int|s|
-        Backend returned a secret that is not a valid \
+        Backend returned an entry that is not a valid \
```
dcastro, line 40: I think this would be more appropriate:

```suggestion
        Backend returned a path segment that is not a valid \
```

lib/Error.hs
```diff
@@ -37,7 +37,7 @@ instance Buildable InternalCommandsError where
   build = \case
     InvalidEntry entry ->
```
dcastro, line 38: Also rename this variable from `entry` to `pathSegment` while we're here.

lib/Backend.hs
```diff
@@ -44,12 +44,12 @@ data BackendEffect m a where
   -- | Overwrites any entry that might already exist at that path.
```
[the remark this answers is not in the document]
sancho20021: How I forgot that  :|

lib/Backend.hs
```diff
@@ -28,10 +28,10 @@ type Effects r = (Member (Embed IO) r, Member (Error CofferError) r)
 class Show a => Backend a where
   _name :: a -> BackendName
   _codec :: Toml.TomlCodec a
-  _writeSecret :: Effects r => a -> Entry -> Sem r ()
-  _readSecret :: Effects r => a -> EntryPath -> Sem r (Maybe Entry)
-  _listSecrets :: Effects r => a -> Path -> Sem r (Maybe [Text])
-  _deleteSecret :: Effects r => a -> EntryPath -> Sem r ()
+  _writeEntry :: Effects r => a -> Entry -> Sem r ()
+  _readEntry :: Effects r => a -> EntryPath -> Sem r (Maybe Entry)
+  _listEntries :: Effects r => a -> Path -> Sem r (Maybe [Text])
```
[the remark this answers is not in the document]
sancho20021: Done

lib/Backend/Commands.hs
```diff
@@ -4,7 +4,7 @@
 
 module Backend.Commands where
```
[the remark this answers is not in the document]
sancho20021: Done

lib/Error.hs
sancho20021: Ok!

lib/Backend/Commands.hs
```diff
@@ -517,26 +518,26 @@ getEntryOrDir backend path =
...
-          forM_ secrets \secret -> do
-            -- We need to find out whether `secret` is a directory name or an entry name.
+          forM_ nodes \node -> do
+            -- We need to find out whether `node` is a directory name or an entry name.
             --
             -- Here, we rely on the fact that Vault returns directory names suffixed by `/`.
             -- and entry names without any suffix.
             --
-            -- If `mkPathSegment` succeeds, then `secret` was not suffixed by `/`,
+            -- If `mkPathSegment` succeeds, then `entry` was not suffixed by `/`,
```
dcastro, line 529:
```suggestion
            -- If `mkPathSegment` succeeds, then `node` was not suffixed by `/`,
```

dcastro (changes requested):

lib/Backend/Commands.hs
sancho20021: Done. I should be more attentive when making quick changes :)

dcastro (approved):

DK318 (approved):

PR #107, opened 2022-05-12
[#106] Backend-agnostic `listDirectoryContents`

### LINKED ISSUE #106: Hide `listEntries`'s implementation details

The `listSecrets` function (soon to be renamed in #100) returns a `[Text]`. These strings can be either entry names or directory names.

Vault will add a `/` suffix to directory names, and `Backend.Commands.getEntryOrDir` relies on this behaviour to decide whether each of those strings is an entry name or a directory name.

However, this behavior is specific to the Vault backend. Other backends might not behave like this.
So we should hide this behavior behind `listSecrets`.

Instead of returning a `[Text]`, it should return a new data structure `DirectoryContents` with two fields: `directoryNames :: [PathSegment]` and `entryNames :: [PathSegment]`.

Also:
* We should add a haddock to `ListSecrets`.
* As part of this task, I think `InvalidEntry` could be moved from `InternalCommandsError` to `VaultError`.

# Acceptance criteria

* `listSecrets` returns an appropriate data structure that doesn't rely on backend-specific details.

DK318:
## Problem 
At this moment the `listDirectoryContents` function returns `[Text]`
and in `getEntryOrDir` we rely on the fact that `vault` adds a '/'
to the directory. But other backends might not behave like this.

## Solution 
Created new data `DirectoryContents` which stores 2 lists:
1. list with directories
2. list with entries
and changed return type from `Maybe [Text]` to `Maybe DirectoryContents`
in `listDirectoryContents` function.

## Related issue(s)

- Fixes #106 

Fixed #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/lib/Backend/Commands.hs
+++ b/lib/Backend/Commands.hs
@@ -11,8 +11,9 @@ import CLI.Types
 import Coffer.Directory (Directory)
 import Coffer.Directory qualified as Dir
 import Coffer.Path
-  (EntryPath, Path, QualifiedPath(QualifiedPath, qpBackendName, qpPath), entryPathName, mkPath,
-  mkPathSegment, pathSegments, unPathSegment)
+  (DirectoryContents(DirectoryContents), EntryPath, Path(Path),
+  QualifiedPath(QualifiedPath, qpBackendName, qpPath), directoryNames, entryNames, entryPathName,
+  pathSegments, unPathSegment)
 import Coffer.Path qualified as Path
 import Coffer.Util (catchAndReturn)
 import Config (Config(backends, mainBackend))
@@ -37,7 +38,7 @@ import Entry
   (Entry, EntryTag, Field, FieldName, FieldVisibility(..), contents, dateModified, fieldContents,
   fields, newEntry, newField, path, visibility)
 import Entry qualified as E
-import Error (CofferError(..), InternalCommandsError(EntryPathDoesntHavePrefix, InvalidEntry))
+import Error (CofferError(..), InternalCommandsError(EntryPathDoesntHavePrefix))
 import GHC.Exts (Down(..), sortWith)
 import Polysemy
 import Polysemy.Error (Error, throw)
@@ -468,7 +469,7 @@ pathIsDirectory :: forall r. Member BackendEffect r => SomeBackend -> EntryPath 
 pathIsDirectory backend entryPath =
   listDirectoryContents backend (Path.entryPathAsPath entryPath) >>= \case
     Nothing -> pure False
-    Just [] -> pure False
+    Just (DirectoryContents [] []) -> pure False
     Just _ -> pure True
 
 -- | Checks if the path points to an existing entry.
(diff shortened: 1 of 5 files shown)


lib/Backend/Vault/Kv.hs
```diff
@@ -284,15 +293,37 @@ kvReadSecret backend path = do
...
-    pure $ response ^. I.ddata . I.unListSecrets
+  result <-
+    embedCatchClientErrorMaybe do
+      response <- listSecrets env (getPathSegments path)
+      let contents = response ^. I.ddata . I.unListSecrets
+      pure $ foldM makeDirectoryAndEntryNames (DirectoryContents [] []) contents
+
+  case result of
+    Nothing -> pure Nothing
+    Just e -> either throw pure e <&> Just
```
dcastro, line 307: Polysemy's [`fromEither`](https://hackage.haskell.org/package/polysemy-1.7.1.0/docs/Polysemy-Error.html#v:fromEither) may help here

```suggestion
    Just e -> Just <$> fromEither e
```

dcastro (approved):
Good job! :+1:

lib/Backend/Vault/Kv.hs

lib/Backend/Commands.hs
```diff
@@ -518,26 +519,17 @@ getEntryOrDir backend path =
...
-                  Just entry -> modify' (Dir.insertEntry entry)
-                  Nothing -> pure ()
-
-              (_, Right subdir) -> go (rootPath <> subdir)
-              _ -> lift $ throw $ InternalCommandsError (InvalidEntry node)
+          forM_ (contents ^. entryNames) \entryName -> do
+            entry <- lift $ readEntry backend (Path.appendEntryName rootPath entryName)
+            case entry of
+              Just entry -> modify' (Dir.insertEntry entry)
+              Nothing -> pure ()
```
sancho20021, line 528: @dcastro If I remember correctly, when we discussed line 528 
```
Nothing -> pure ()
```
you said it is not correct to just ignore unexpected `Nothing`. 

Maybe we can throw some *internalError*, cause `contents ^. entryNames` returned an entryName that couldn't be read properly by `readEntry`, which is unexpected.

sancho20021 (changes requested):
LGTM, except that `pure ()` situation needs to be revisited :) (Maybe add TODO or fix it in this PR)

lib/Backend/Commands.hs
DK318: I just copy-pasted that code). Maybe we should really throw something here

dcastro: Yeah I remember saying that. I've been thinking about it a bit more, and maybe `pure ()` is right after all :thinking: 

So, we're traversing a directory, `listDirectoryContents` gave us a list of entry names, and then `readEntry` returned 404 for one of those entry names.

That, to me, suggests that some other user deleted an entry from `vault` (or whatever backend) _while_ we were traversing the directory.

So, there's two things we can do here:
1. ignore this entry that has been concurrently deleted
2. throw and crash

(2) would imply that `coffer` can only traverse a vault directory if _no one_ else is messing with it. I don't think we want to do this, we want to allow multiple people to interact with vault at the same time. So ignoring the entry is the sensible thing to do here, IMO.

dcastro: I do think we should add a comment explaining what this branch means.

``` 
Nothing ->
  -- This entry has been concurrently deleted (e.g. by some other user) _while_ we're traversing the directory.
  -- We should just ignore it.
  pure ()
```

lib/Error.hs
```diff
@@ -30,17 +30,10 @@ class (Buildable err) => BackendError err
 
 -- | Internal errors that can be thrown if backend is in illegal state.
 data InternalCommandsError
-  = InvalidEntry Text
-  | EntryPathDoesntHavePrefix EntryPath Path
+  = EntryPathDoesntHavePrefix EntryPath Path
 
 instance Buildable InternalCommandsError where
   build = \case
```
Kariel-Myrr, line 36: Seems we can get rid of `\case`, because `InternalCommandError` have only one way now. 
But it will add work to person how will add new `InternalCommandError` and make code more inconsistent(look at `CofferError` instance).

Kariel-Myrr (approved):
LGTM 🌚

lib/Backend/Commands.hs
DK318: Done

sancho20021 (approved):
Lgtm

PR #108, opened 2022-05-16
[#98] One style errors

### LINKED ISSUE #98: Use one style for showing parsing errors

Currently CLI uses different styles to signal syntax errors.

## Examples:
- **sort**
```
>>> coffer find --sort BAD_FORMAT
option --sort: Invalid sort format: "BAD_FORMAT".
Expected format is:
 * to sort by entry name: 'name:<direction>',
 * to sort by the entry's last modified date: 'date:<direction>',
 * to sort by a field's contents: '<fieldname>:contents:<direction>',
 * to sort by a field's last modified date: '<fieldname>:date:<direction>'.
Direction can be 'asc' or 'desc'.
Examples: 'name:desc', 'password:date:asc'.
...
```
- **filter**
```
>>> coffer find --filter BAD_FORMAT
option --filter: Invalid filter format: "BAD_FORMAT".
Expected format is:
 * to filter by entry name: 'name~<substring>',
 * to filter by entry's last modified date: 'date<op><date>',
 * to filter by a field's contents: '<fieldname>:contents~<substring>',
 * to filter by a field's last modified date: '<fieldname>:date<op><date>'.
<op> can be '<=', '>=', '<', '>', or '='.
<date> can be 'YYYY', 'YYYY-MM', 'YYYY-MM-DD', or 'YYYY-MM-DD HH:MM:SS'.
Examples: 'name~vault', 'date<2020-02', 'url:contents~google.com', 'pw:date<2020-02'.

Parser error:
1:11:
  |
1 | BAD_FORMAT
  |           ^
unexpected end of input
expecting ':' or fieldname
...
```
In case of `sort` custom messages are shown, while in `filter`, `megaparsec`'s error messages appear. To keep things consistent, we can use `megaparsec` for both.

# Acceptance criteria

Rewrite parsing of `--sort` option using `megaparsec` instead of `Text.splitOn`.

Kariel-Myrr:
## Description

Problem : Parse error messages in `find --sort` and `find --find` had different styles

Solution : rewrote `sort` parsing to Megaparsec instead of `splitOn(":")`

## Related issue(s)

Fixed #98 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/lib/CLI/Parser.hs
+++ b/lib/CLI/Parser.hs
@@ -13,7 +13,6 @@ import BackendName (BackendName, newBackendName)
 import CLI.Types
 import Coffer.Path (EntryPath, Path, QualifiedPath(QualifiedPath), mkEntryPath, mkPath)
 import Coffer.Util (MParser)
-import Control.Arrow ((>>>))
 import Control.Monad (guard, void)
 import Data.Bifunctor (first)
 import Data.Char qualified as Char
@@ -421,33 +420,13 @@ readFieldInfo = do
 readSort :: ReadM (Sort, Direction)
 readSort = do
   eitherReader \input ->
-    case T.splitOn ":" (T.pack input) of
-      [means, direction] -> do
-        direction' <- readDirection direction
-        case means of
-          "name" -> pure (SortByEntryName, direction')
-          "date" -> pure (SortByEntryDate, direction')
-          _ -> Left [int|s|
-            Invalid sort: #{show means}.
-            Choose one of: 'name', 'date'.
-
-            #{show expectedSortFormat}
-          |]
-      [fieldName, means, direction] -> do
-        fieldName' <- readFieldName' fieldName
-        direction' <- readDirection direction
-        case means of
-          "contents" -> pure (SortByFieldContents fieldName', direction')
-          "date" -> pure (SortByFieldDate fieldName', direction')
-          _ -> Left [int|s|
-            Invalid sort: #{show means}.
-            Choose one of: 'contents', 'date'.
-
-            #{show expectedSortFormat}
-          |]
-      _ -> Left [int|s|
+    P.parse (parseSort <* P.eof) "" (T.pack input) & first \err ->
+      [int|s|
         Invalid sort format: #{show input}.
         #{show expectedSortFormat}
+
+        Parser error:
+        #{P.errorBundlePretty err}
       |]
 
 expectedSortFormat :: Pretty.Doc
@@ -460,13 +439,6 @@ expectedSortFormat = Pretty.vsep
   , "Direction can be 'asc' or 'desc'."
   , "Examples: 'name:desc', 'password:date:asc'."
   ]
-
-readDirection :: Text -> Either String Direction
-readDirection =
-  T.unpack >>> readSum "direction"
-    [ ("asc", Asc)
-    , ("desc", Desc)
-    ]
 
 readFilter :: ReadM Filter
 readFilter = do
@@ -567,6 +539,38 @@ parseFieldNameWhile whileCond = do
 -- | Parse the rest of the input as a field content.
 parseFieldContentsEof :: MParser FieldContents
 parseFieldContentsEof = FieldContents . T.pack <$> P.manyTill P.anySingle P.eof
+
+parseSort :: MParser (Sort, Direction)
+parseSort = do
+  sort <- parseSortMeans
+  void $ P.char ':'
+  direction <- parseSortDirection
+  return (sort, direction)
+
+parseSortDirection :: MParser Direction
+parseSortDirection =
+      P.string "asc" $> Asc
+  <|> P.string "desc" $> Desc
+
+parseSortMeans :: MParser Sort
+parseSortMeans =
+  try parseSortMeansByFieldContentsOrDate
+  <|> parseSortMeansByNameOrDate
+
+parseSortMeansByFieldContentsOrDate :: MParser Sort
+parseSortMeansByFieldContentsOrDate = do
+  fieldName <- parseFieldNameWhile (/= ':')
+  void $ P.char ':'
+  P.choice @[] $
+    [ (SortByFieldDate fieldName)     <$ (P.string "date")
+    , (SortByFieldContents fieldName) <$ (P.string "contents")
+    ]
+
+parseSortMeansByNameOrDate :: MParser Sort
+parseSortMeansByNameOrDate = P.choice @[] $
+  [ SortByEntryName <$ P.string "name"
+  , SortByEntryDate <$ (P.string "date")
+  ]
 
 ----------------------------------------------------------------------------
 -- Utils
--- a/tests/golden/find-command/find-command.bats
+++ b/tests/golden/find-command/find-command.bats
@@ -131,6 +131,54 @@ EOF
     a - [2000-01-01 01:01:01]
       sortfield: c [2000-01-01 01:01:01]
 EOF
+
+
+}
+
+@test "sort on field with name 'date' and 'name'" {
+
+  coffer create /secrets/a
+  coffer create /secrets/c
+  coffer create /secrets/b
+
+  coffer set-field /secrets/a name name1
+  coffer set-field /secrets/c name name2
+  coffer set-field /secrets/b name name3
+
+  run cleanOutput coffer find / --sort name:contents:desc
+
+  assert_success
+  assert_output - <<EOF
+/
+  secrets/
+    b - [2000-01-01 01:01:01]
+      name: name3 [2000-01-01 01:01:01]
+    c - [2000-01-01 01:01:01]
+      name: name2 [2000-01-01 01:01:01]
+    a - [2000-01-01 01:01:01]
+      name: name1 [2000-01-01 01:01:01]
+EOF
+
+  coffer set-field /secrets/a date date1
+  coffer set-field /secrets/c date date2
+  coffer set-field /secrets/b date date3
+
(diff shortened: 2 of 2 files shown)


lib/CLI/Parser.hs
```diff
@@ -568,6 +549,41 @@ parseFieldNameWhile whileCond = do
 parseFieldContentsEof :: MParser FieldContents
 parseFieldContentsEof = FieldContents . T.pack <$> P.manyTill P.anySingle P.eof
 
+
```
DK318, line 552: Same

lib/CLI/Parser.hs
```diff
@@ -568,6 +549,41 @@ parseFieldNameWhile whileCond = do
...
+  try parseSortMeansByNameOrDate
+  <|> parseSortMeansByFieldName
+
+parseSortMeansByFieldName :: MParser Sort
+parseSortMeansByFieldName = do
+  fieldName <- parseFieldNameWhile (/= ':')
+  void $ P.char ':'
+  P.choice (
+    [ (SortByFieldDate fieldName)     <$ (P.string "date")
+    , (SortByFieldContents fieldName) <$ (P.string "contents") ] :: [MParser Sort] )
```
DK318, line 577: You can avoid using parentheses like this
```suggestion
  P.choice @[] @_ @Sort $
    [ (SortByFieldDate fieldName)     <$ (P.string "date")
    , (SortByFieldContents fieldName) <$ (P.string "contents")
    ]
```
Btw, in list declarations we put a trailing `]` (see [code-style](https://github.com/serokell/coffer/blob/main/docs/code-style.md#list-declarations))

lib/CLI/Parser.hs
```diff
@@ -568,6 +549,41 @@ parseFieldNameWhile whileCond = do
...
+
+
+parseSortMeansByNameOrDate :: MParser Sort
+parseSortMeansByNameOrDate = P.choice (
+  [ SortByEntryName <$ P.string "name"
+  , SortByEntryDate <$ (P.string "date") ] :: [MParser Sort] )
```
DK318, line 583: Same here

DK318 (changes requested):

DK318:
And, please, when you start merging it, clean up your `git` history

DK318:
Please, don't resolve conversations on your own. They are resolving by the same person who opened it

lib/CLI/Parser.hs
```diff
@@ -568,6 +546,39 @@ parseFieldNameWhile whileCond = do
...
 parseFieldContentsEof = FieldContents . T.pack <$> P.manyTill P.anySingle P.eof
 
+parseSort :: MParser (Sort, Direction)
+parseSort = do
+  sort <- parseSortMeans
+  void $ P.char ':'
+  direction <- parseSortDirection
+  return (sort, direction)
+
+
```
DK318, line 556: Another extra line :smile:

DK318 (approved):
LGTM overall

lib/CLI/Parser.hs
```diff
@@ -568,6 +546,39 @@ parseFieldNameWhile whileCond = do
...
+parseSortDirection :: MParser Direction
+parseSortDirection =
+  try (P.string "asc" $> Asc)
```
sancho20021, line 559:
```suggestion
  P.string "asc" $> Asc
```
I predict that this `try` is not needed. Looks like we either need `try` before both alternatives or we don't need them at all.

sancho20021 (approved):
Ok!

lib/CLI/Parser.hs
```diff
@@ -568,6 +546,39 @@ parseFieldNameWhile whileCond = do
...
+  <|> P.string "desc" $> Desc
+
+parseSortMeans :: MParser Sort
+parseSortMeans =
+  try parseSortMeansByNameOrDate
+  <|> parseSortMeansByFieldName
+
+parseSortMeansByFieldName :: MParser Sort
```
dcastro, line 567:
```suggestion
parseSortMeansByFieldContentsOrDate :: MParser Sort
```

lib/CLI/Parser.hs
dcastro: Yeah I don't think it's needed either.

lib/CLI/Parser.hs
```diff
@@ -568,6 +546,39 @@ parseFieldNameWhile whileCond = do
...
+
+
+parseSortDirection :: MParser Direction
+parseSortDirection =
+  try (P.string "asc" $> Asc)
+  <|> P.string "desc" $> Desc
+
+parseSortMeans :: MParser Sort
+parseSortMeans =
+  try parseSortMeansByNameOrDate
```
dcastro, line 564: We should have a test for checking that this `try` here backtracks as intended.

Essentially, we want to make sure that when a have a field named `name`, then this is interpreted correctly by the parser:

```
--sort name:contents:asc
       ^^^^
          | This `name` does not mean "the entry's name", it means "a field named 'name'"
```

I have a feeling this test is not going to pass though, we might have to rework a few things.

dcastro: Indeed, it won't pass:


```
$ coffer find / --sort name:contents:asc

option --sort: Invalid sort format: "name:contents:asc".
Expected format is:
 * to sort by entry name: 'name:<direction>',
 * to sort by the entry's last modified date: 'date:<direction>',
 * to sort by a field's contents: '<fieldname>:contents:<direction>',
 * to sort by a field's last modified date: '<fieldname>:date:<direction>'.
Direction can be 'asc' or 'desc'.
Examples: 'name:desc', 'password:date:asc'.

Parser error:
1:6:
  |
1 | name:contents:asc
  |      ^^^^
unexpected "cont"
expecting "asc" or "desc"
```

The `parseSortMeansByNameOrDate` parses succeeded in matching on `name`, it didn't backtrack, and then it proceeded to `parseSortDirection` and failed.

dcastro (changes requested):

lib/CLI/Parser.hs
Kariel-Myrr: I think it could be solved by changing the order of parsing means. So we first will parse field name -> contents/date, and if not we will parse name as keyword
But we will get err from `--sort name:contents` not for `name:direction` but for expecting `name:contents:-missing direction-`

Kariel-Myrr: Also we can prevent farther problems with amount of arguments by adding followedBy ':' in `parseSortMeansByFieldContentsOrDate` part. So it will be successfully ended if only we have another argument after

tests/golden/find-command/find-command.bats
```diff
@@ -131,6 +131,20 @@ EOF
...
+  assert_output - <<EOF
+/
+  secrets/
+    a - [2000-01-01 01:01:01]
+      sortfield: c [2000-01-01 01:01:01]
+    b - [2000-01-01 01:01:01]
+      sortfield: b [2000-01-01 01:01:01]
+    c - [2000-01-01 01:01:01]
+      sortfield: a [2000-01-01 01:01:01]
+EOF
```
dcastro, line 147: Nice :+1: 

Since this is testing a separate edge case, let's put this assertion in its own dedicated test, e.g. `test "sort on field with name 'date' and 'name'"`

And also add a similar test for a field named `date`

lib/CLI/Parser.hs
dcastro: I'm not sure I understand. The error messages for `name:contents` and `name:contents:` seem reasonable to me:

```
$ coffer find / --sort name:contents 
...

Parser error:
1:14:
  |
1 | name:contents
  |              ^
unexpected end of input
expecting ':'
```

```
$ coffer find / --sort name:contents:
...

Parser error:
1:15:
  |
1 | name:contents:
  |               ^
unexpected end of input
expecting "asc" or "desc"
```

lib/CLI/Parser.hs
```diff
@@ -11,9 +11,14 @@ module CLI.Parser
 
 import BackendName (BackendName, newBackendName)
 import CLI.Types
+  (Command(..), CopyOptions(CopyOptions), CreateOptions(CreateOptions),
+  DeleteFieldOptions(DeleteFieldOptions), DeleteOptions(DeleteOptions), Direction(..),
+  FieldInfo(FieldInfo), Filter(..), FilterField(FilterFieldByContents, FilterFieldByDate),
+  FilterOp(..), FindOptions(FindOptions), Options(Options), RenameOptions(RenameOptions),
+  SetFieldOptions(SetFieldOptions), SomeCommand(..), Sort(..), TagOptions(TagOptions),
+  ViewOptions(ViewOptions), parseFilterDate)
```
dcastro, line 19: No need for explicit import list here

dcastro (approved):
LGTM :+1: 

Just one last very small change, and then please clean up the commit history ^^

lib/CLI/Parser.hs
Kariel-Myrr: But stylish-haskell made it :>

dcastro: Are you sure? AFAIK stylish-haskell doesn't turn implicit import lists into explicit.

I think you may have accidentally clicked the "code lens" that HLS adds above the import:

image

If you click that code lens (in grey), HLS will make the import list explicit.

Can you please revert this?

dcastro:
Small typo in the first commit's message:

> So error diff betw one command

lib/CLI/Parser.hs
DK318: We are using specific `stylish-haskell` version (0.13.0.0 AFAIR)

PR #110, opened 2022-05-17
[#82] Pretty `ToJSON` instances

### LINKED ISSUE #82: Prettify JSON instances
This issue is blocked by #37. It would probably be a good idea to do #83 first.

The JSON representation of the web API leaks a lot of internal details about our data structures. For example, the `/view` endpoint leaks a lot of haskell field names such as `dEntries` and `unEntryPath` and haskell constructor names such as `VRDirectory`:

```json
{
  "contents": {
    "dEntries": [],
    "dSubdirs": {
      "dir": {
        "dEntries": [
          {
            "ePath": {
              "unEntryPath": [
                "dir",
                "entry3"
              ]
            },
            "eFields": {
              "user": {
                "fValue": "diogo",
                "fVisibility": "public",
                "fDateModified": "2022-04-28T17:21:56.774631254Z"
              }
            },
            "eDateModified": "2022-04-28T17:21:56.774631254Z",
            "eMasterField": null,
            "eTags": []
          }
        ],
        "dSubdirs": {}
      }
    }
  },
  "tag": "VRDirectory"
}
```

In some cases, it's just a matter of using the right functions for renaming fields (e.g. with the `aeson-casing` package):

```hs
deriveJSON (aesonPrefix camelCase) ''TypeName
```

In other cases, the instances should be handcrafted. E.g., the entry path should be rendered as `"/dir/entry3"` rather than:
```json
"unEntryPath": [
  "dir",
  "entry3"
]
```

# Acceptance criteria

* The `ToJSON` instances don't print any haskell specific details.
* I'm not sure if the `FromJSON` instances are needed or used at this point. If they are needed, `ToJSON` and `FromJSON` should roundtrip.
* For handcrafted instances, we should have roundtripping tests.

DK318:
## Problem 
At this moment `ToJSON` instances leak a lot of internal info.
Some of them look very ugly (e.g. `EntryPath`).

## Solution 
Used `deriveToJSON (aesonPrefix camelCase)` to derive some instances.
`QualifiedPath` `ToJSON` instance is handcrafted. Also removed unnecessary
`FromJSON` instances.

## Related issue(s)

- Fixes #82 

Fixed #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/lib/CLI/Types.hs
+++ b/lib/CLI/Types.hs
@@ -5,11 +5,14 @@
 module CLI.Types where
 
 import Coffer.Directory (Directory)
-import Coffer.Path (EntryPath, Path, QualifiedPath)
+import Coffer.Path (EntryPath, Path, QualifiedPath(qpPath))
 import Coffer.Util (MParser)
 import Control.Applicative (Alternative(some), optional)
+import Control.Lens hiding (noneOf)
 import Control.Monad (guard, void)
 import Data.Aeson hiding ((<?>))
+import Data.Aeson.QQ (aesonQQ)
+import Data.Aeson.Types (emptyObject)
 import Data.Bifunctor (first)
 import Data.Char qualified as Char
 import Data.Fixed (Pico)
@@ -64,35 +67,31 @@ data ViewResult
   | VRPathNotFound (QualifiedPath Path)
   | VRDirectoryNoFieldMatch (QualifiedPath Path) FieldName
   | VREntryNoFieldMatch (QualifiedPath EntryPath) FieldName
-  deriving stock (Show, Generic)
-  deriving anyclass (FromJSON, ToJSON)
+  deriving stock (Show)
 
 data CreateError
   = CEParentDirectoryIsEntry (QualifiedPath EntryPath, QualifiedPath EntryPath)
   | CEDestinationIsDirectory (QualifiedPath EntryPath)
   | CEEntryAlreadyExists (QualifiedPath EntryPath)
   deriving stock (Show, Generic)
-  deriving anyclass (FromJSON, ToJSON)
+  deriving anyclass (ToJSON)
 
 data CreateResult
   = CRSuccess (QualifiedPath EntryPath)
   | CRCreateError CreateError
-  deriving stock (Show, Generic)
-  deriving anyclass (FromJSON, ToJSON)
+  deriving stock (Show)
 
 data SetFieldResult
   = SFRSuccess FieldName (QualifiedPath Entry)
   | SFREntryNotFound (QualifiedPath EntryPath)
   | SFRMissingFieldContents FieldName (QualifiedPath EntryPath)
-  deriving stock (Show, Generic)
-  deriving anyclass (FromJSON, ToJSON)
+  deriving stock (Show)
 
 data DeleteFieldResult
   = DFRSuccess FieldName (QualifiedPath EntryPath)
   | DFREntryNotFound (QualifiedPath EntryPath)
   | DFRFieldNotFound FieldName
-  deriving stock (Show, Generic)
-  deriving anyclass (FromJSON, ToJSON)
+  deriving stock (Show)
 
 type RenameResult = CopyResult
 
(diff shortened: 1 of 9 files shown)


lib/CLI/Types.hs
```diff
@@ -327,6 +324,102 @@ instance FromHttpApiData (FieldName, FilterField) where
               return (field, FilterFieldByDate op date)
           ]
 
+-- These @ToJSON@ instances are for success results only.
```
sancho20021, line 327: If error produces empty object json, do we really need `"result": "ok"` at all?

lib/Web/Server.hs
```diff
@@ -44,12 +44,13 @@ import Servant.Server
 data CofferServerError = CofferServerError
   { cseError :: Text
   , cseCode :: Int
+  , cseResult :: Text
```
sancho20021, line 47: Hmm, I don't understand the purpose of this field. Could you please explain it's function?

lib/CLI/Types.hs
DK318: At this moment all errors are produced in `Server` module. It's expected that if we call `toJSON` on results, the result is successful. I made the "result" field to understand whether our action was successful or not.

lib/Web/Server.hs
DK318: Check my answer above

lib/CLI/Types.hs
sancho20021: Oh, I see. But isn't responding with error code enough (which you already do) to understand that operation was unsuccessful? (I have little experience in web programming so my opinion can be groundless haha 😂)

lib/Coffer/Path.hs
```diff
@@ -196,8 +200,16 @@ data QualifiedPath path = QualifiedPath
...
+  deriving stock (Show, Eq, Functor)
+
+instance (Buildable path) => ToJSON (QualifiedPath path) where
+  toJSON qPath = do
+    let value = String $ pretty qPath
+    [aesonQQ|
+      {
+        "path": #{value}
+      }
+    |]
```
dcastro, line 212: Why do we wrap the path string in a json object with 1 field? Wouldn't `toJSON = String . pretty`, like you did for `EntryPath`, work?

lib/CLI/Types.hs
```diff
@@ -149,14 +146,14 @@ data SetFieldOptions = SetFieldOptions
...
-  deriving anyclass (FromJSON, ToJSON)
+  deriving anyclass (ToJSON)
 
 data DeleteFieldOptions = DeleteFieldOptions
   { dfoQPath :: QualifiedPath EntryPath
   , dfoFieldName :: FieldName
   }
   deriving stock (Show, Generic)
-  deriving anyclass (FromJSON, ToJSON)
+  deriving anyclass (ToJSON)
```
dcastro, line 156: Looks like most of these `ToJSON` instances aren't needed either, might as well delete them

dcastro (changes requested):

lib/CLI/Types.hs
dcastro: >  But isn't responding with error code enough (which you already do) to understand that operation was unsuccessful?

Yeah, I agree. A consumer of our API should look at the status code - if it's a 2xx, they should decode the response body to a directory/entry/whatever, and if it's any other status code they should expect the response body to have the format `{ "error": "", "code": n }`

---

A separate, but related issue, is that we now have instances like this:

```hs
instance ToJSON CreateResult where
  toJSON = \case
    CRSuccess entryPath ->
      [aesonQQ|
        {
          "contents": #{entryPath},
          "result": "ok"
        }
      |]
    _ -> emptyObject
```

Where `toJSON` is safe to use with _some values_ (`CRSuccess`), but when used with other values (`CRCreateError`) it'll produce meaningless output, which kinda goes ahead haskell's mantra.

DK318: I don't know how to write instances for errors. I depend on functions from `Coffer.PrettyPrint` but I can't import them because of cyclic dependencies.

lib/Web/Server.hs
DK318: Removed anyway

lib/Coffer/Path.hs
DK318: Done

lib/CLI/Types.hs
DK318: Done

dcastro (approved):

sancho20021 (approved):

PR #112, opened 2022-05-20
[#111] Correct endpoints results in Web API

### LINKED ISSUE #111: Correct endpoints results in Web API

At this moment in Web API in all endpoints (except for `find`) we return the corresponding result from `CLI.Types`. But it seems not correct, because these results contain both success and errors. And API should return only success results (error results return in corresponding handlers. See #99).

Endpoints that create resources should return `Entry`. For e.g. `/create` endpoint should return `Entry` instead of `CreateResult`.

Endpoints that modify resources should return that modified resource. For e.g. `/set-field` endpoint should return `Entry` and `/copy` endpoint should return `Directory`.

# Acceptance criteria
All Web API endpoints return `Entry` or `Directory` results.

DK318:
## Problem 
At this moment in Web API in all endpoints (except for `find`)
we return the corresponding result from `CLI.Types`.

But it seems not correct, because these results contain
both success and errors. And API should return only success results
(error results return in corresponding handlers).

## Solution 
Changed endpoints results to `Entry | Directory`. Some changes
demanded changes in `CLI.Types` (e.g. returning `Entry` instead of `EntryPath`).

## Related issue(s)

- Fixes #111 

Fixed #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/lib/Web/API.hs
+++ b/lib/Web/API.hs
@@ -19,29 +19,28 @@ type API
   :> "api" :> "v1" :> "content" :>
     ( "view"
       :> RequiredParam "path"  (QualifiedPath Path)
-      :> OptionalParam "field" FieldName
-      :> Get '[JSON] ViewResult
+      :> Get '[JSON] Directory
 
     :<|> "create"
       :> RequiredParam "path" (QualifiedPath EntryPath)
       :> QueryFlag     "force"
       :> ReqBody '[JSON] NewEntry
-      :> Post '[JSON] CreateResult
+      :> Post '[JSON] Entry
 
     :<|> "set-field"
       :> RequiredParam "path" (QualifiedPath EntryPath)
       :> RequiredParam "field" FieldName
       :>
-        (    "private" :> Post '[JSON] SetFieldResult
-        :<|> "public"  :> Post '[JSON] SetFieldResult
+        (    "private" :> Post '[JSON] Entry
+        :<|> "public"  :> Post '[JSON] Entry
         :<|> ReqBody '[JSON] (Maybe FieldContents)
-          :> Post '[JSON] SetFieldResult
+          :> Post '[JSON] Entry
         )
 
     :<|> "delete-field"
       :> RequiredParam "path" (QualifiedPath EntryPath)
       :> RequiredParam "field" FieldName
-      :> Delete '[JSON] DeleteFieldResult
+      :> Delete '[JSON] Entry
 
     :<|> "find"
       :> OptionalParam  "path" (QualifiedPath Path)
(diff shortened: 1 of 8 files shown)


lib/Web/API.hs
```diff
@@ -55,27 +54,27 @@ type API
...
       :> QueryFlag     "force"
-      :> Post '[JSON] CopyResult
+      :> Post '[JSON] [(EntryPath, EntryPath)]
 
     :<|> "delete"
       :> QueryFlag     "dry-run"
       :> RequiredParam "path" (QualifiedPath Path)
       :> QueryFlag     "recursive"
-      :> Delete '[JSON] DeleteResult
+      :> Delete '[JSON] Directory
```
dcastro, line 70: `DELETE` endpoints usually return a 200 with an empty response body when successful.

```suggestion
      :> DeleteNoContent
```

(I guess `/delete-field` is a bit different because it doesn't delete _the whole_ resource, just a part of it, so I guess in that particular case it makes sense to return the `Entry`)

lib/Web/Server.hs
```diff
@@ -69,11 +70,12 @@ throwCofferServerErrors err codesAndMsgs = do
...
-handleViewResult :: ViewResult -> Handler ViewResult
+handleViewResult :: ViewResult -> Handler Directory
 handleViewResult = \case
-  res@VRDirectory{} -> pure res
-  res@VREntry{} -> pure res
-  res@VRField{} -> pure res
+  VRDirectory dir -> pure dir
+  VREntry entry -> pure $ singleton entry
+  VRField{} ->
+    throwError $ err500 { errBody = "Unexpected VRField" }
```
dcastro, line 78: Same suggestion as in [this comment](https://github.com/serokell/coffer/pull/99#discussion_r878303809), use `throwCofferServerError` with an error code of 0.

dcastro (changes requested):

lib/Web/API.hs
DK318: Done

lib/Web/Server.hs
DK318: Done

dcastro (approved):

sancho20021 (commented):
Isn't that a little bit confusing that we put `Entry` (an entire object) inside `QualifiedPath` (a path to the object)?

sancho20021:
> > Isn't that a little bit confusing that we put Entry (an entire object) inside QualifiedPath (a path to the object)?
> 
> A little, but it reduces code in `copy/rename` functions

Oh I now saw that we are going to get rid of `QualifiedPath` in WebAPI due to [Issue 113](https://github.com/serokell/coffer/issues/113). Soo it will either be rewritten, right?

sancho20021 (approved):

PR #117, opened 2022-06-01
[#89] Reuse existing megaparsec parsers

### LINKED ISSUE #89: Reuse existing megaparsec parsers

The Web API reimplements some parsing logic to parse things like `Filter` (see the `instance FromHttpApiData Filter`).

We should reuse the same parsers we used in the CLI modules.

Additionally, the logic to parse a `QualifiedPath` is duplicated in `FromHttpApiData (QualifiedPath path)` and in `CLI.Parser.readQualifiedEntryPath`. We should try to have a single `mkQualifiedPath` function that is reused in the `FromHttpApiData` instance and in the CLI parser.

# Acceptance criteria

* The parsing logic is not duplicated.

Kariel-Myrr:
## Problem:
We had a lot of repeating code in `FromHttpApiData` instances with `CLI.Parse`.

## Solution:
That was fixed.
Basically a lot of low parsing logic moved from `CLI.Parse` to `CLI.Types` and `Coffer.Path`.
It was necessary to avoid cyclic dependencies.

## Related issue(s)

- Resolves #89 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/lib/CLI/Parser.hs
+++ b/lib/CLI/Parser.hs
@@ -484,7 +484,7 @@ expectedFilterFormat = Pretty.vsep
 
 parseFilterOp :: MParser FilterOp
 parseFilterOp =
-  asum @[]
+  P.choice @[]
     [ P.string ">=" $> OpGTE
     , P.string "<=" $> OpLTE
     , P.char '>' $> OpGT
--- a/lib/CLI/Parser.hs
+++ b/lib/CLI/Parser.hs
@@ -13,10 +13,8 @@ import BackendName (BackendName, newBackendName)
 import CLI.Types
 import Coffer.Path (EntryPath, Path, QualifiedPath(QualifiedPath), mkEntryPath, mkPath)
 import Coffer.Util (MParser)
-import Control.Monad (guard, void)
+import Control.Monad (void)
 import Data.Bifunctor (first)
-import Data.Char qualified as Char
-import Data.Foldable (asum)
 import Data.Function ((&))
 import Data.Functor (($>), (<&>))
 import Data.List qualified as List
@@ -26,8 +24,6 @@ import Data.Set qualified as Set
 import Data.Text (Text)
 import Data.Text qualified as T
 import Entry
-  (EntryTag, FieldContents(FieldContents), FieldName, FieldVisibility(..), newEntryTag,
-  newFieldName)
 import Fmt (pretty)
 import Options.Applicative
 import Options.Applicative.Help.Pretty qualified as Pretty
@@ -357,18 +353,6 @@ readFieldVisibility =
     , ("private", Private)
     ]
 
-readFieldName :: ReadM FieldName
-readFieldName = str >>= toReader . readFieldName'
-
-readFieldName' :: Text -> Either String FieldName
-readFieldName' input = do
-  case newFieldName input of
-    Right tag -> pure tag
-    Left err -> Left $ unlines
-      [ "Invalid field name: " <> show input <> "."
-      , pretty err
-      ]
-
 readQualifiedEntryPath :: ReadM (QualifiedPath EntryPath)
 readQualifiedEntryPath = do
   eitherReader \input ->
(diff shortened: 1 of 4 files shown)


lib/Coffer/Path.hs
```diff
@@ -221,16 +224,31 @@ instance (Buildable path) => Buildable (QualifiedPath path) where
...
+  parseUrlPiece = mkQualifiedPath
+    (\_ -> "Invalid qualified path format. Expected [BACKENDNAME#]PATH")
+    parseUrlPiece
+
+mkQualifiedPath
+  :: (Text -> Text)
+  -> (Text -> Either Text path)
+  -> (Text -> Either Text (QualifiedPath path))
+mkQualifiedPath errorMessage mkPath text = case T.splitOn "#" text of
+      [backendNameStr, pathStr] -> do
```
dcastro, line 236: Excessive indentation here, let's keep it to 2 spaces

lib/Coffer/Path.hs
```diff
@@ -221,16 +224,31 @@ instance (Buildable path) => Buildable (QualifiedPath path) where
...
+        backendName <- readBackendName' backendNameStr
```
dcastro, line 237: We can just use `newBackendName` directly

```suggestion
        backendName <- newBackendName backendNameStr
```

lib/Coffer/Path.hs
```diff
@@ -221,16 +224,31 @@ instance (Buildable path) => Buildable (QualifiedPath path) where
...
+        path <- mkPath pathStr
         pure $ QualifiedPath (Just backendName) path
-    | otherwise =
-        Left "Invalid qualified path format. Expected [BACKENDNAME#]PATH"
+      [pathStr] -> do
+        path <- mkPath pathStr
+        pure $ QualifiedPath Nothing path
+      _ -> Left $ errorMessage text
```
dcastro, line 243: Instead of taking in an `errorMessage :: Text -> Text` parameter, let's just return an error message explaining what went wrong, like what the other `mk*` constructors do.

So, what went wrong in the `_` branch? Either `T.splitOn "#"` returned a list with 0 elements or a list with >2 elements.

I don't think `splitOn` ever returns an empty list, at least according to my experiments, so we can rule that out. That means it must have returned a list with >2 elements. Which means the string contained more than one '#' characters, so let's return an error message that says that.

lib/CLI/Parser.hs
```diff
@@ -318,9 +309,9 @@ tagOptions =
 -- Common
 ----------------------------------------------------------------------------
 
-readPath' :: Text -> Either String Path
+readPath' :: Text -> Either Text Path
```
dcastro, line 312: Oh, a bit unrelated, but can you please delete `_readEntryPath` while you're here? It's not used anywhere

dcastro, line 312: We should do the same to `readEntryPath'` for consistency, change `Either String` to `Either Text`

lib/Coffer/Path.hs
Kariel-Myrr: Done

lib/Coffer/Path.hs
Kariel-Myrr: Done

lib/CLI/Parser.hs
Kariel-Myrr: Done

Kariel-Myrr: Done

dcastro (approved):

tests/golden/common/common.bats
```diff
@@ -43,7 +43,6 @@ EOF
```
DK318, line 43: Imo it would be better if we split fixing this test case in another commit.

DK318 (approved):
LGTM overall

tests/golden/common/common.bats
Kariel-Myrr: I could, but it would mean on one stage our test's won't  pass. Is it acceptable? And this changes directly connected to code changes

PR #127, opened 2022-07-21
[#65] Make path's charset rules backend specific

### LINKED ISSUE #65: Make path's charset rules backend-specific

Currently coffer restricts all Entry paths to `['a'..'z'] ++ ['A'..'Z'] ++ ['0'..'9'] ++ "-_"` while Unix/Linux allows for many more. This may cause issues with file based backends like pass. Rather than cheking globally, if a backend has limitations it should enforce them at its boundary.

# Acceptance criteria

Either allow more chars or close as `wontfix`.

Kariel-Myrr:
### Motivation
We wanted to validate paths not on Path lvl, but on Backend lvl (So each backend validates paths)

### Solution
- Remove check for allowed symbols
- Added check for not containing '#'
- Added new Effect of `Backend` - `validatePath`
- Added validation for paths in `Backend.Commands`
     So it's now cmd responsibility to check if path is valid lies on cmd
- Updated error msg - now they contain info about specific backend
- Added new backend specific tests

## Related issue(s)

Fixed #65

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/lib/Backend.hs
+++ b/lib/Backend.hs
@@ -9,7 +9,7 @@ module Backend
 where
 
 import BackendName (BackendName)
-import Coffer.Path (DirectoryContents, EntryPath, Path)
+import Coffer.Path (DirectoryContents, EntryPath, HasPathSegments, Path)
 import Data.Aeson qualified as A
 import Entry (Entry)
 import Error (CofferError)
@@ -26,3 +26,5 @@ class (Show a, A.FromJSON a) => Backend a where
   _readEntry :: Effects r => a -> EntryPath -> Sem r (Maybe Entry)
   _listDirectoryContents :: Effects r => a -> Path -> Sem r (Maybe DirectoryContents)
   _deleteEntry :: Effects r => a -> EntryPath -> Sem r ()
+  _validatePath :: Effects r => HasPathSegments s segments
+    => a -> s -> Sem r ()
--- a/lib/Backend/Commands.hs
+++ b/lib/Backend/Commands.hs
@@ -4,7 +4,8 @@
 
 module Backend.Commands where
 
-import BackendEffect (BackendEffect, deleteEntry, listDirectoryContents, readEntry, writeEntry)
+import BackendEffect
+  (BackendEffect, deleteEntry, listDirectoryContents, readEntry, validatePath, writeEntry)
 import BackendName (BackendName)
 import Backends (SomeBackend)
 import CLI.Types
@@ -61,8 +62,9 @@ runCommand config = \case
 viewCmd
   :: (Members '[BackendEffect, Error CofferError, Error ViewResult] r)
   => Config -> ViewOptions -> Sem r ViewResult
-viewCmd config (ViewOptions qPath@(QualifiedPath backendNameMb _) fieldNameMb) = do
+viewCmd config (ViewOptions qPath@(QualifiedPath backendNameMb path) fieldNameMb) = do
   backend <- getBackend config backendNameMb
+  validatePath backend path
   getEntryOrDirThrow backend VRPathNotFound qPath >>= \case
     Right dir -> do
       case fieldNameMb of
@@ -97,6 +99,7 @@ createCmd
   (CreateOptions qEntryPath@(QualifiedPath backendNameMb entryPath) _edit force tags fields privateFields)
     = do
   backend <- getBackend config backendNameMb
+  validatePath backend entryPath
   nowUtc <- embed getCurrentTime
   let
     mkField :: FieldVisibility -> FieldInfo -> (FieldName, Field)
@@ -123,6 +126,7 @@ setFieldCmd
   (SetFieldOptions qEntryPath@(QualifiedPath backendNameMb entryPath) fieldName fieldContentsMb visibilityMb)
     = do
   backend <- getBackend config backendNameMb
+  validatePath backend entryPath
   readEntry backend entryPath >>= \case
     Nothing -> do
       pure $ SFREntryNotFound qEntryPath
@@ -162,6 +166,7 @@ deleteFieldCmd
   => Config -> DeleteFieldOptions -> Sem r DeleteFieldResult
 deleteFieldCmd config (DeleteFieldOptions qPath@(QualifiedPath backendNameMb path) fieldName) = do
   backend <- getBackend config backendNameMb
+  validatePath backend path
   readEntry backend path >>= \case
     Nothing -> pure $ DFREntryNotFound qPath
     Just entry -> do
@@ -210,6 +215,7 @@ findCmd config (FindOptions qPathMb textMb sortMb filters) = do
             FilterFieldByDate op date -> matchDate op date (field ^. dateModified)
 
   let path = maybe mempty qpPath qPathMb
+  validatePath backend path
   dir <- getEntryOrDir backend path <&> \case
     Just (Right dir) -> dir
     Just (Left entry) -> Dir.singleton entry
@@ -276,13 +282,15 @@ renameCmd
   config
   (RenameOptions
     dryRun
-    oldQPath@(QualifiedPath oldBackendNameMb _)
-    newQPath@(QualifiedPath newBackendNameMb _)
+    oldQPath@(QualifiedPath oldBackendNameMb oldPath)
+    newQPath@(QualifiedPath newBackendNameMb newPath)
     force
   )
     = do
   oldBackend <- getBackend config oldBackendNameMb
   newBackend <- getBackend config newBackendNameMb
+  validatePath oldBackend oldPath
+  validatePath newBackend newPath
   operations <- buildCopyOperations oldBackend newBackend oldQPath newQPath force
 
   unless dryRun do
@@ -395,13 +403,15 @@ copyCmd
   config
   (CopyOptions
     dryRun
-    oldQPath@(QualifiedPath oldBackendNameMb _)
-    newQPath@(QualifiedPath newBackendNameMb _)
+    oldQPath@(QualifiedPath oldBackendNameMb oldPath)
+    newQPath@(QualifiedPath newBackendNameMb newPath)
     force
   )
     = do
   oldBackend <- getBackend config oldBackendNameMb
   newBackend <- getBackend config newBackendNameMb
+  validatePath oldBackend oldPath
+  validatePath newBackend newPath
   operations <- buildCopyOperations oldBackend newBackend oldQPath newQPath force
 
   unless dryRun do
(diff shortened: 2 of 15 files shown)


lib/Backend/Commands.hs
```diff
@@ -436,6 +446,7 @@ tagCmd
   => Config -> TagOptions -> Sem r TagResult
 tagCmd config (TagOptions qEntryPath@(QualifiedPath backendNameMb entryPath) tag delete) = do
   backend <- getBackend config backendNameMb
+  validatePath backend entryPath
```
dcastro, line 449: Forgot to validate the path in `deleteCmd`

tests/golden/vault/common.bats
```diff
@@ -0,0 +1,35 @@
...
+load '../helpers/bats-assert/load'
+load '../helpers'
+
+@test "vault bad path" {
+  run coffer view "$(echo -e "bad\npath")"
+
+  assert_failure
+  assert_output --partial - <<EOF
+Internal backend error:
+Backend returned a path segment that is not a valid entry or directory name.
```
dcastro, line 17: > Backend returned a path segment that is not a valid entry or directory name

The message seems off. This path was returned by the backend; it was entered by the user.

tests/golden/vault/common.bats
```diff
@@ -0,0 +1,35 @@
...
+load '../helpers/bats-support/load'
+load '../helpers/bats-assert/load'
+load '../helpers'
+
+@test "vault bad path" {
+  run coffer view "$(echo -e "bad\npath")"
+
+  assert_failure
+  assert_output --partial - <<EOF
+Internal backend error:
```
dcastro, line 16: > Internal backend error:

This isn't an "internal" error either.

Internal errors are (most likely) bugs in our code or in any internal component, like vault. But, in this case, the error was due to bad data entered by the user, so it's not an internal error.

lib/Backends.hs
```diff
@@ -49,7 +70,7 @@ backendPackedCodec = Toml.Codec input output
         Just t -> do
           case Toml.backward Toml._Text t >>= supportedBackends of
             Right c -> c toml
-            Left e -> Failure [ Toml.BiMapError "type" e ]
+            Left _ -> Failure [ Toml.BiMapError "type" (Toml.ArbitraryError "Unknown backend type") ]
```
dcastro, line 73: Why not return `e`?

lib/Backends.hs
```diff
@@ -60,5 +81,6 @@ backendPackedCodec = Toml.Codec input output
 
 supportedBackends
   :: Text -> Either Toml.TomlBiMapError (Toml.TomlEnv SomeBackend)
-supportedBackends "vault-kv" = Right $ fmap SomeBackend . Toml.codecRead (_codec @VaultKvBackend)
-supportedBackends _ = Left (Toml.ArbitraryError "Unknown backend type")
+supportedBackends backend = case HS.lookup (T.unpack backend) supportedBackendsMap of
+  Just x -> Right (bToml x)
+  Nothing ->  Left (Toml.ArbitraryError "Unknown backend type")
```
dcastro, line 86: I think at this point it makes sense to inline `supportedBackends` into `backendPackedCodec`.

(And, while we're here, rename `backendPackedCodec` to `someBackendCodec`, seems like a more appropriate name)

lib/Backends.hs
```diff
@@ -28,13 +32,30 @@ data SomeBackend where
...
+  }
+
+supportedBackendsMap :: HS.HashMap String SupportedBackend
+supportedBackendsMap = HS.fromList [("vault-kv", vaultKVSup)]
+  where
+    vaultKVSup = SupportedBackend
+      "vault-kv"
+      ((fmap SomeBackend) . (A.parseJSON @VaultKvBackend))
+      (fmap SomeBackend . Toml.codecRead (_codec @VaultKvBackend))
+      "Vault Kv paths can contain only the following characters: [a-zA-Z0-9] and symbols '-', '_'."
```
dcastro, line 49: Here as well: we should use `kvPathSegmentAllowedCharacters` to generate this message

lib/CLI/Parser.hs
```diff
@@ -443,6 +449,9 @@ expectedFilterFormat = Pretty.vsep
 -- Utils
 ----------------------------------------------------------------------------
 
+getBackendsHelpMsg :: (SupportedBackend -> Pretty.Doc) -> [Pretty.Doc]
```
dcastro, line 452: `getBackendsHelpMsg`'s first argument is always `bPathHelpMsg`, so it looks like it doesn't really need the `SupportedBackend -> Pretty.Doc` parameter.

tests/golden/common/common.bats
```diff
@@ -53,11 +53,6 @@ EOF
 
   assert_failure
   assert_output --partial - <<EOF
-Invalid qualified entry path format: "back#/path#\n\nsmth".
-Expected format is: [<backend-name>#]<entry-path>.
-<backend-name> can be a string of the following characters: [a-zA-Z0-9] and symbols '-', '_', ';'.
-Examples: 'vault_kv-backend#secrets/google', 'my/passwords/entry'.
-
```
dcastro, line 69: I think we should keep these lines in the test's expected output.

dcastro (changes requested):

lib/CLI/Parser.hs
Kariel-Myrr: I wanted to keep space for other help msgs
But, ok, will get rid of it

lib/Backends.hs
Kariel-Myrr: Yeah, true. It was a mix of old code

lib/Backend.hs
```diff
@@ -26,3 +27,5 @@ class (Show a, A.FromJSON a) => Backend a where
   _readEntry :: Effects r => a -> EntryPath -> Sem r (Maybe Entry)
   _listDirectoryContents :: Effects r => a -> Path -> Sem r (Maybe DirectoryContents)
   _deleteEntry :: Effects r => a -> EntryPath -> Sem r ()
+  _validatePath :: Effects r => (HasPathSegments s segments, Each segments segments PathSegment PathSegment)
```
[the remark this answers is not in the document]
Kariel-Myrr: I've made... smth
It works, but I'm not sure that it's intended way
I'm understanding why we need Iso & each. But do we still need `HasPathSegments` layer?
Seems I just had to add Each constraint to `HasPathSegments`. Because we always want segments contain `PathSegment`

lib/Backend/Vault/Kv.hs
```diff
@@ -111,7 +112,9 @@ instance Buildable VaultError where
       [int|s|
         Backend returned a path segment that is not a valid \
         entry or directory name.
-        Got: '#{pathSegment}'.
+        Path segments for Vault KV can only contain the following characters:
+          'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_'
```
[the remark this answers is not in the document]
Kariel-Myrr: done

lib/Backend/Commands.hs
Kariel-Myrr: done

lib/Backends.hs
Kariel-Myrr: done
I haven't get rid of supportedBackends completely 
it was the best way to save initial structure

lib/Backends.hs
Kariel-Myrr: done

tests/golden/vault/common.bats
Kariel-Myrr: done

tests/golden/vault/common.bats
Kariel-Myrr: done

tests/golden/common/common.bats
Kariel-Myrr: done

lib/Backends.hs
Kariel-Myrr: done

lib/CLI/Parser.hs
Kariel-Myrr: done
So it was too small to be a function, so I left a pure map

lib/Error.hs
```diff
@@ -24,6 +24,7 @@ data CofferError where
   BadFieldNameError :: BadFieldName -> CofferError
   BadMasterFieldName :: Text -> BadFieldName -> CofferError
   BadEntryTagError :: BadEntryTag -> CofferError
+  InvalidPathSegment :: BackendError err => err -> CofferError
```
dcastro, line 27: Making `InvalidPathSegment` a wrapper for `BackendError err => err` can lead to invalid states. E.g., it allows us to do this:

```
throw $ InvalidPathSegment $ ServantError ...
```

We should restrict this constructor like this:

```suggestion
InvalidPathSegment
  :: PathSegment
  -> Text -- ^ Backend-specific error message
  -> CofferError
```

In this case, the second argument `Text` would be this section of the error message:

```
Path segments for Vault KV can only contain the following characters:
'#{kvPathSegmentAllowedCharacters}'
```

dcastro: And then delete the `VaultInvalidPathSegment` constructor

dcastro (changes requested):

lib/Backend.hs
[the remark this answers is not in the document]
dcastro: As discussed in the call, instead of having a `SuperPathSegmented` class, let's add an `Each` constraint to the `HasPathSegments` class:

```hs
class Each segments segments PathSegment PathSegment => HasPathSegments s segments | s -> segments where
  pathSegments :: Iso' s segments
```

lib/Error.hs
Kariel-Myrr: done

lib/Backend.hs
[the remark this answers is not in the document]
Kariel-Myrr: done

tests/golden/vault/common.bats
```diff
@@ -14,11 +14,11 @@ load '../helpers'
   assert_failure
   assert_output --partial - <<EOF
 Invalid path segment for target backend:
+Got: 'bad
+path'.
```
dcastro, line 18: Oh, we want to escape characters in invalid paths (e.g., we should show `\n` instead of ` `).

The reason for this is that some invalid characters might be hard to spot _unless_ they're escaped.

For example, at the moment, if I use a path with a ["zero-width space"](https://unicode-table.com/en/200B/), it won't be visible in the error message:

```
$ coffer create "$(echo "bad\xE2\x80\x8Bpath")"
Invalid path segment for target backend:
Got: 'badpath'.
      ^^^^^^^ seems like a valid path, but it's not, there's a zero-width space character in the middle.
```


If we escape all characters, then we would see:

```
$ coffer create "$(echo "bad\xE2\x80\x8Bpath")"
Invalid path segment for target backend:
Got: "bad\8203path".
```

To escape the characters, we can use `#s` instead of `#` to tell the `nyan-interpolation` to use the `Show Text` instance instead of the `Buildable Text` instance:

```hs
-- before
Got: '#{segment}'.

-- after
Got: #s{unPathSegment segment}.
```

Kariel-Myrr: oh, that's nice)
done

dcastro (approved):
LGTM :smile:

Issue #119, opened 2022-06-02
Refactor `/set-field` endpoint


dcastro:
# Clarification and motivation

The `set-field` endpoints are defined as:

```hs
    :<|> "set-field"
      :> RequiredParam "path" (QualifiedPath EntryPath)
      :> RequiredParam "field" FieldName
      :>
        (    "private" :> Post '[JSON] Entry
        :<|> "public"  :> Post '[JSON] Entry
        :<|> ReqBody '[JSON] (Maybe FieldContents)
          :> Post '[JSON] Entry
        )
```

In other words, we have 3 distinct endpoints:

* `/set-field` with an optional `Maybe FieldContents` body
* `/set-field/private` with no body
* `/set-field/public` with no body

There are multiple issues with this interface:

1. The body of the `/set-field` endpoint is `Maybe FieldContents`, which suggests it's okay to call it with `null` in the body. But actually, the endpoint fails when we do that:
    ```
    $ curl -s -D /dev/stderr -H 'token: root' -X POST -H 'Content-Type: application/json' \
      'localhost:8081/api/v1/content/set-field?path=/entry&field=fff' \
      -d 'null' | jq

    HTTP/1.1 400 Bad Request
    Transfer-Encoding: chunked
    Date: Thu, 02 Jun 2022 15:43:29 GMT
    Server: Warp/3.3.18
    Content-Type: application/json;charset=utf-8

    [
      {
        "error": "The entry at '/entry' does not yet have a field 'fff'.\nIn order to create a new field, please include 'FIELDCONTENTS' in the body.",
        "code": 301
      }
    ]
    ```
1. To create a private field, we need two separate calls:
  * `POST /set-field` with the field contents in the body
  * `POST /set-field/private`

These issues can be solved by having a single endpoint that accepts an optional `?visibility=public/private` query parameter **and** an optional `FieldContents`.

We should also be able to inline `handleSetFieldResult` as part of this issue.

# Acceptance criteria

We have a single `/set-field` endpoint

sancho20021:
@dcastro by optional FieldContents you mean `ReqBody '[JSON] FieldContents`? But if it is null, the problem remains the same:
- do nothing if a field already exists
- error if a field doesn't exist

The combination *visibility=null && fieldcontents=null* is a bit strange, I think that is the main issue, which won't be resolved with your suggestion. Or probably I misunderstood something.

dcastro:
> the problem remains the same:

[...]

Issue #105, opened 2022-05-11
Write tests for the Web API


dcastro:
# Clarification and motivation

This issue is blocked by #104.

I don't think we need to test every edge case for the Web API as we did for the CLI. Since they use the same core modules, that seems like duplicated effort.

We should treat it as "just another interface" for coffer's core, and test it as such: 
* make sure all inputs (URI path, query params, headers, etc) are correctly parsed and processed
* make sure all outputs (response body, headers, status codes, etc) are correct

Testing the output of error scenarios does involve testing a lot of edge cases, but perhaps not as many as we did in the CLI, I hope.

We want these tests to serve as documentation. So let's not derive a servant client, let's use a lower level abstraction. And let's not use coffer's core data types/encoders/decoders, let's use json literals (e.g. with [aesonQQ](https://hackage.haskell.org/package/aeson-2.0.3.0/docs/Data-Aeson-QQ-Simple.html#v:aesonQQ))

# Acceptance criteria

* We have a reasonable test coverage of the Web API

PR #128, opened 2022-07-24
[#119] Change set-field Web-API

sancho20021:
## Description

Problem: `set-field` route was used both for changing visibility
and setting new fields. This resulted in additional handling of the
situation where both field contents and visibility option are missing.

Solution: require contents for setting field and allow changing
visibility for convenience. Add `set-field-visibility` request
which modifies only visibility.

### Question?
Currently in `set-field-visibility` I pass the value in the body of the request.
Maybe it would be better to pass it in the request string itself:

```
set-field-visibility/entry-path/field/private
```

I don't know.

## Related issue(s)
Fixed part of #119. Refactor of CLI API still needed.

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/tests/server-integration/SetField/SetFieldTest.hs
+++ b/tests/server-integration/SetField/SetFieldTest.hs
@@ -6,13 +6,11 @@ module SetField.SetFieldTest where
 
 import Control.Exception (try)
 import Control.Lens
-import Data.Aeson
 import Data.Aeson.Lens
 import Data.Aeson.QQ.Simple (aesonQQ)
-import Data.Text (Text)
 import Data.Time
 import Network.HTTP.Req
-import Network.HTTP.Types (status400, status404)
+import Network.HTTP.Types (status404)
 import Test.Tasty.HUnit
 import Utils
 
@@ -20,7 +18,7 @@ unit_set_new_field :: IO ()
 unit_set_new_field = cofferTest do
   createEntry "entry"
 
-  response <- setField "entry" "newField" "contents"
+  response <- setField "entry" "newField" Nothing "contents"
   response @=
     [aesonQQ|
       {
@@ -41,26 +39,16 @@ unit_set_new_field = cofferTest do
 unit_set_field_private :: IO ()
 unit_set_field_private = cofferTest do
   createEntry "entry"
-  setField "entry" "public-field" "contents"
-  response <-
-    executeCommand
-      POST
-      ["set-field", "private"]
-      NoReqBody
-      (jsonResponse @Value)
-      ( mconcat
-          [ "path" =: ("entry" :: Text)
-          , "field" =: ("public-field" :: Text)
-          ]
-      )
+
+  response <- setField "entry" "private-field" (Just False) "contents"
 
   let fieldMb =
         responseBody response
           & scrubDates
-          & findField "public-field"
+          & findField "private-field"
 
   case fieldMb of
-    Nothing -> assertFailure $ "Expected field \"public-field\", found nothing"
+    Nothing -> assertFailure $ "Expected field \"private-field\", found nothing"
     Just value ->
       value @?=
         [aesonQQ|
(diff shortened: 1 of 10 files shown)


Kariel-Myrr:
Seems in issue description field visibility should be a query parametr
> These issues can be solved by having a single endpoint that accepts an optional ?visibility=public/private query parameter and an optional FieldContents.

sancho20021:
> Seems in issue description field visibility should be a query parametr

Yes but that was an initial proposal, we later decided to split our queries into `set-field` with indeed an optional `visibility` parameter and `set-field-visibility`.

I don't want the `set-field-visibility` to take visibility as a parameter because it would require repeating the word `visibility` two times:
`/set-field-visibility/?visibility=private`

Kariel-Myrr:
Ok, IMO it would be more logical to make it query. JSON looks like overshot. And seems nothing chain us to JSON format

sancho20021:
> Ok, IMO it would be more logical to make it query. JSON looks like overshot. And seems nothing chain us to JSON format

Yeah, will redo.

Btw what is the expected behavior when changing the visibility of a non-existent field? It's easier to implement that we do nothing and return OK. @dcastro what do you think

sancho20021:
@Kariel-Myrr I moved visibility from JSON body to the query string. Now a request looks like
```
/set-field-visibility/public
```

tests/server-integration/SetField/SetFieldTest.hs
```diff
@@ -116,40 +98,16 @@ unit_set_field_entry_not_found = cofferTest do
...
-      [aesonQQ|
-        [
-          {
-            "error": "The entry at '/entry' does not yet have a field 'field'.\nIn order to create a new field, please include 'FIELDCONTENTS' in the body.",
-            "code": 301
-          }
-        ]
-      |]
+-- unit_set_field_missing_field_contents :: IO ()
+-- TODO: add test that expects an error in case of missing contents
```
Kariel-Myrr, line 102: Seems as unsolved TODO

tests/server-integration/SetField/SetFieldTest.hs
```diff
@@ -158,3 +116,51 @@ unit_set_field_updates_modification_date = cofferTest do
...
   assertBool "Dates are not equal"
     $ responseBody response ^?! key "fields" . key "field" . key "dateModified" . _JSON == newModifiedDate
+
+unit_set_field_visibility :: IO ()
+unit_set_field_visibility = cofferTest do
+  createEntry "dir/entry"
+  void $ setField "dir/entry" "private-field" Nothing "contents"
+  void $ setField "dir/entry" "public-field" Nothing "contents"
+  changeFieldVisibility "dir/entry" "private-field" "private"
+  changeFieldVisibility "dir/entry" "public-field" "public"
```
Kariel-Myrr, line 126: Why there is two checks in one test?

If we're expecting to fail at once = one way (only private/public)
If it could fail on one, but not on second variant seems as two separate tests

I know that they're just doing the same thing, but we aren't short on test space)

tests/server-integration/Utils.hs
```diff
@@ -102,8 +102,8 @@ scrubDates =
 cofferTest :: IO () -> IO ()
 cofferTest test = deleteRecords >> test
 
-setField :: Text -> Text -> Text -> IO (JsonResponse Value)
-setField path name contents =
+setField :: Text -> Text -> Maybe Bool -> Text -> IO (JsonResponse Value)
```
Kariel-Myrr, line 105: Why not straight `Maybe Visibility`

Kariel-Myrr (commented):
LGTM

tests/server-integration/Utils.hs
sancho20021: Because in #105 it says:
> And let's not use coffer's core data types/encoders/decoders, let's use json literals (e.g. with [aesonQQ](https://hackage.haskell.org/package/aeson-2.0.3.0/docs/Data-Aeson-QQ-Simple.html#v:aesonQQ))

We don't import anything from coffer core so just for consistency I don't import `Visibility` too

sancho20021: I think I will change `Bool` to `Text` to unify the style with function `changeFieldVisibility`

tests/server-integration/SetField/SetFieldTest.hs
sancho20021: Fixed :+1:

tests/server-integration/SetField/SetFieldTest.hs
sancho20021: Resolved now, you can check how I did it, I'm not sure it is the proper way

sancho20021:
This PR is closed due to existing https://github.com/serokell/coffer/pull/130.

PR #130, opened 2022-08-01
[#119] Refactor `set-field` in CLI

### LINKED ISSUE #119: Refactor `/set-field` endpoint

sancho20021:
## Description

**Problem:** currently `set-field` is used both for changing visibility
and setting new fields. This results in additional handling of the
situation where both field contents and visibility option are missing.

**Solution:** require contents for setting field and allow changing
visibility for convenience. Add `set-field-visibility` command
which modifies only visibility.

## Related issue(s)

Fixes #119 

## Blocks
I created this branch from [sancho20021/#119-refactor-set-field-endpoint](https://github.com/serokell/coffer/tree/sancho20021/%23119-refactor-set-field-endpoint) so I assume this PR should not be merged until https://github.com/serokell/coffer/pull/128 is merged. 

Please correct me If I should have picked main branch as the target instead of [sancho20021/#119-refactor-set-field-endpoint](https://github.com/serokell/coffer/tree/sancho20021/%23119-refactor-set-field-endpoint).

## TODO

Change Web API status codes documentation according to this PR. I can do it in this PR if necessary.

- Tests
        silently reappearing again.

- Documentation
    - Haddock


lib/Backend/Commands.hs
```diff
@@ -136,26 +141,60 @@ setFieldCmd
...
+    updateOrInsertFieldFull :: UTCTime -> Maybe Field -> Sem r (Maybe Field)
+    updateOrInsertFieldFull nowUtc =
+      pure . Just .
+      (visibility %~ \current -> fromMaybe current visibilityMb) .
+      updateOrInsertField nowUtc
 
-    updateOrInsertField :: UTCTime -> Maybe Field -> Sem r (Maybe Field)
-    updateOrInsertField nowUtc = \case
+    updateOrInsertField :: UTCTime -> Maybe Field -> Field
+    updateOrInsertField nowUtc fieldMb = case fieldMb of
```
Kariel-Myrr, line 153: \case?)

lib/Web/Server.hs
```diff
@@ -125,23 +134,29 @@ reportErrors io = do
     Right (Right a) -> do
       return a
 
+-- setFieldVisibility
```
Kariel-Myrr, line 137: Looks like leftover

Kariel-Myrr (commented):
LGTM
(not touching Web part)

Kariel-Myrr (changes requested):
We've added new cmd. Seems we have to add it to README

app/cli/Main.hs
```diff
@@ -85,9 +86,15 @@ main = do
       SomeCommand cmd@CmdSetField{} -> do
         runCommand config cmd >>= \case
           res@SFREntryNotFound{} -> printError $ buildSetFieldResult CLI res
-          res@SFRMissingFieldContents{} -> printError $ buildSetFieldResult CLI res
           res@SFRSuccess{} -> printSuccess $ buildSetFieldResult CLI res
 
+
```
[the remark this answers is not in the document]
sancho20021: Fixed

lib/Backend/Commands.hs
sancho20021: Yes, thanks

lib/Web/Server.hs
sancho20021: 100%. Sorry

sancho20021:
> We've added new cmd. Seems we have to add it to README

True. Fixed. And also documentation about status codes

PR #132, opened 2022-10-25
[Chore] Update xrefcheck to v0.2.2

Sereja313:
## Description

Update xrefcheck to v0.2.2.
## Related issue(s)

None

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -58,7 +58,7 @@ steps:
       - "result/bin/coffer"
 
   - label: xrefcheck
-    command: nix run -f ci.nix xrefcheck -c xrefcheck --no-progress -m full --ignored tests/golden/helpers
+    command: nix run -f ci.nix xrefcheck -c xrefcheck
 
   - label: REUSE lint
     command: nix run -f ci.nix pkgs.reuse -c reuse lint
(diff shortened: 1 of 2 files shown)


dcastro:
Same as in tzbot, let's delete `--no-progress -m full`

.buildkite/pipeline.yml
```diff
@@ -58,7 +58,7 @@ steps:
       - "result/bin/coffer"
 
   - label: xrefcheck
-    command: nix run -f ci.nix xrefcheck -c xrefcheck --no-progress -m full --ignored tests/golden/helpers
+    command: nix run -f ci.nix xrefcheck -c xrefcheck --ignored tests/golden/helpers
```
dcastro, line 61: Should this be turned into a glob pattern? `tests/golden/helpers/**/*`

dcastro: Oh wait, I don't think we even need to ignore this anymore.

Previosly, we were using `--ignored` here to ignore markdown files contained inside the `bats` / `bats-assert` / `bats-support` submodules, e.g.:

* tests/golden/helpers/bats/README.md
* tests/golden/helpers/bats-support/README.md
* tests/golden/helpers/bats-assert/README.md

But, because these are submodules, those files are not really tracked in our repository. `git ls-files` only returns their directories, not the files inside:

```
$ git ls-files
...
tests/golden/helpers/bats
tests/golden/helpers/bats-assert
tests/golden/helpers/bats-support
...
```

And since 0.2.2 uses `git ls-files`, we don't need to use `--ignored` for this anymore.

dcastro (approved):

